PR #1, opened 2015-01-30
First implementation

pilhuhn:
This is a first implementation of HAWKULAR-1, which uses the embedded H2 as backend.
It is not meant as the final implementation or api, but rather as a start of a building block that 
can be used in the larger Hawkular scenario.


pilhuhn:
I will send an updated (and squashed) PR

PR #3, opened 2015-02-04
HWKINVENT-5 update parent pom and logging prefix

pilhuhn:


pilhuhn:
The travis failure comes from the fact that Travis does

```
mvn install
mvn test
```

Which triggers a compiler bug on the second "install" which implicitly happens during mvn test.

For me that happens with the bsc processor, as also locally with the mvn compiler plugin. We need to investigate that on a more global scale.

metlos:
Let's keep this unmerged for a couple of days just to have a reference of the problem.

metlos:
Btw. wouldn't "mvn clean test" help here?

ppalaga:
@pilhuhn is it necessary to let Travis to perform both `mvn install` and `mvn test` ? Is it not enough to run just one of them in travis.yml's `script` section? Here in hk build tools, we have just a dummy `mvn -version` in travis.yml's `install`: https://github.com/hawkular/hawkular-build-tools/blob/master/.travis.yml#L7 
It is only in `script` where the real stuff happens: https://github.com/hawkular/hawkular-build-tools/blob/master/.travis.yml#L10

pilhuhn:
Travis first tries to compile/build the whole project and then test.
Which makes sense, as you don't want to prevent compiling just because the
first test falls flat on the nose.

Also as developer you want to be able to run mvn test twice in a row without cleaning
in between

> Am 04.02.2015 um 18:36 schrieb Peter Palaga notifications@github.com:
> 
> @pilhuhn is it necessary to let Travis to perform both mvn install and mvn test ? Is it not enough to run just one of them in travis.yml's script section? Here in hk build tools, we have just a dummy mvn -version in travis.yml's install: https://github.com/hawkular/hawkular-build-tools/blob/master/.travis.yml#L7 
> It is only in script where the real stuff happens: https://github.com/hawkular/hawkular-build-tools/blob/master/.travis.yml#L10
> 
> —
> Reply to this email directly or view it on GitHub.

## 

Heiko Rupp   hwr@pilhuhn.de
Blog: http://pilhuhn.blogspot.com   @pilhuhn

PR #5, opened 2015-02-05
POM-work to prevent compiler bug with Annotation processor.

pilhuhn:
This requires parent pom v5, which is not yet on central, so the first CI run will fail.

If you have the jboss nexus in ~/.m2/settings.xml you can build it , as the jboss releases repo already has v5


pilhuhn:
I'll close this one and redo

PR #9, opened 2015-02-10
Add Cors filter to the project

stefannegrea:
Also for testing purposes add a ping handler that replies with that current date.


pilhuhn:
I guess that is just a copy from the metrics one?

PR #11, opened 2015-02-13
Updating REST interface for clearer hierarchy

vrockai:
I've unified the "resources" and "resource" end-points + used PUT where id was specified in URL.


metlos:
Looks fine to me, but I'll let @pilhuhn chime in, just in case this would break his integration code unnecessarily (at this moment).

PR #12, opened 2015-02-13
Notifier

pilhuhn:


metlos:
This contains stuff that I think is already merged into master under different commit hashes.

Also, you have to promise to me this will disappear after the MVP ;-) Architecturally, notifications don't belong into inventory proper, I believe.

pilhuhn:
> Am 13.02.2015 um 17:22 schrieb Lukas Krejci notifications@github.com:
> 
> This contains stuff that I think is already merged into master under different commit hashes.

I think git is clever enough to sense this ( I may be wrong though).

> Also, you have to promise to me this will disappear after the MVP ;-) Architecturally, notifications don't belong into inventory proper, I believe.

Not sure I understand the last sentence. What I've implemented is q&d and can be changed going forward. 
In general I think a lot of components need to send around notifications to (yet unknown) consumers instead 
of consumers like the pinger constantly polling e.g. inventory for news.
Such a consumer can also be the UI with 3rd party plugins that we will never hear about.

metlos:
> > Also, you have to promise to me this will disappear after the MVP ;-)
> > Architecturally, notifications don't belong into inventory proper, I
> > believe.
> > Not sure I understand the last sentence. What I've implemented is q&d and
> > can be changed going forward. In general I think a lot of components need
> > to send around notifications to (yet unknown) consumers instead of
> > consumers like the pinger constantly polling e.g. inventory for news. Such
> > a consumer can also be the UI with 3rd party plugins that we will never
> > hear about.

I may have missed some important development in the design but I was under the 
impression that notifications are going to be part of the glue so that the 
components don't have to depend on a message bus (which drags quite a few deps 
with it).

As far as I recall each project is supposed to have a bus-based component but 
this should not be part of the core codebase but rather to be used as an 
"enabling feature" when the component is to be integrated in a bigger picture 
like Hawkular "whole". The component should be usable without it.

With the changes you made, you're coupling inventory to the bus very tightly. 
I understand that - we need to make stuff appear to work quickly. I just 
needed to make the remark that it is not how it should be done in the end.

pilhuhn:
> Am 13.02.2015 um 17:58 schrieb Lukas Krejci notifications@github.com:
> 
> I may have missed some important development in the design but I was under the 
> impression that notifications are going to be part of the glue so that the 

Yes and no. There are certainly components (like metrics) where this is more important.

> I understand that - we need to make stuff appear to work quickly. I just 
> needed to make the remark that it is not how it should be done in the end.

I was just not sure in your previous remark if you did not want to have notifications at all or if the current implementation is bothering you.

pilhuhn:
The build now passes as the dependencies are available in Nexus now.

PR #20, opened 2015-02-23
Adding couple of tests, testing the Related.{by|asTargetBy} filtering.

jkremser:


api/src/main/java/org/hawkular/inventory/api/MultipleEntityBrowser.java
```diff
@@ -23,7 +23,7 @@
  * @author Lukas Krejci
  * @since 1.0
  */
-interface MultipleEntityBrowser<Entity> {
+public interface MultipleEntityBrowser<Entity> {
```
metlos: This interface was supposed to be an internal detail of the API not really exposed to the end-users, but I can see a use for it the way you're using it in the tests.

Still I am not sure if it is advisable to expose it. While we can add a javadoc saying that this is not to be implemented on its own that is not the same as outright forbidding it by access rules.

jkremser: It allows for more general form of use. Types are good, but something more abstract what returns the entitites might be useful also for the API itself. I agree, the word `Browser` might expose some internal terminology, what about something more general name like `HasEntities` with a single method `entities()` ?

metlos: On Monday, February 23, 2015 06:16:57 Jirka Kremser wrote:

> > @@ -23,7 +23,7 @@
> > - @author Lukas Krejci
> > - @since 1.0
> >   */
> > 
> > -interface MultipleEntityBrowser<Entity> {
> > +public interface MultipleEntityBrowser<Entity> {
> 
> It allows for more general form of use. Types are good, but something more
> abstract what returns the entitites might be useful also for the API
> itself. I agree, the word `Browser` might expose some internal terminology,
> what about something more general name like `HasEntities` with a single
> method `entities()` ?

Well, `HasEntities` is not the precise term to use here, because the method 
really just "materializes" the inventory traversal into a set of entities.

Trying hard to think of a meaningful name but I've failed so far.

metlos: What about `ResolvableToMany` for `MultipleEntityBrowser` and `ResolvableToSingle` for `SingleEntityBrowser`?

jkremser: `ResolvableTo{Many|Single}` sounds much better than `MultipleEntityBrowser`.

metlos: Good, so let's go with that then.

Could you rename the two interfaces and make them public, too? I'd merge this PR once that's done.

jkremser: done

PR #23, opened 2015-02-23
Support for quickly steering to an entity during inventory traversal.

metlos:


api/src/main/java/org/hawkular/inventory/api/Inventory.java
```diff
@@ -26,4 +34,83 @@
...
     Tenants.ReadWrite tenants();
+
+    /**
+     * Quickly navigate to the provided tenant.
+     *
+     * @param tenant the tenant to steer to.
+     * @return the access interface to the tenant
+     * @throws EntityNotFoundException if the tenant is not found
+     */
+    default Tenants.Single steerTo(Tenant tenant) throws EntityNotFoundException {
```
jkremser: Why default methods in the interface instead of old fashion abstract class? To me the default methods are good for adding new methods for class that already extends something without breaking the compatibility (e.g. `List.stream()`). 
Now, we have the luxury that no one is using the API so we can break whatever we want. I'd perhaps save the default methods for the future.

metlos: Well, default methods have 2 uses.
1. the one you mention - adding new methods to ifaces in a backwards compatible manner
2. providing default behavior (essentially replacing abstract base impls - if the default behavior is stateless). See for example [here](https://github.com/tinkerpop/tinkerpop3/blob/master/gremlin-core/src/main/java/com/tinkerpop/gremlin/structure/Element.java#L49).

These are obviously the second use case.

If you really don't want them to be default though, I would not make these methods part of this interface at all, because they are essentially helper methods. Instead I'd add a utility class that would offer an API like:

```
Steer.in(inventory).to(environment)...
```

jkremser: Helper class with static methods could be perhaps better here. To me `inventory.steerTo(..)` evokes that I somehow changed the internal state of the `inventory` instance, but that's not the case, it only "prepares" the query and returns it as a Single or Multiple object, waiting for the entity/entities call (or other calls).

What about "look up" or "find"
`Find.in(inv).environment(env)`
because neither looking up, nor finding can change the state. But again, here it's not clear that the result wont be the environment, but something that needs to be materialized.

I am perhaps over-complicating things, here.

metlos: I agree that these should really be out of the `Inventory` interface. Now we just need to find a good name.

I chose "steer" or "navigate" because to me that stresses the act of quering more than the actual retrieval of the entities (which "find" or "look up" would).

what about:

```
Inspect.in(inventory).environment(env)...
```

inspecting something in inventory doesn't evoke state change nor retrieval, IMHO...

metlos: maybe the above could even be folded into the `Inventory` interface after all:

```
inventory.inspect(env)
```

does that still look good to you as it does to me? :wink: 

Of course to make sense, the `inspect` methods should be default on the `Inventory` interface so that impls don't have to bother with it (if they don't want to).

PR #24, opened 2015-02-23
HWKINVENT-15 Relationship service

jkremser:
First chunk. To be continued...
- [x] `get().entity()`
- [x] `named().entities()`
- [ ] `getAll(Filter...)`
- [x] possible consequent calls like `tenants()`, `environments()`, etc. respecting all the previous filters and introducing a new one on the based on the target type of the relationship
- [x] consider introducing a `RelationshipBrowser` (it would need some refactoring, because `Relationship` doesn't extend `Entity`).
- [ ] all the unit tests

### DIFF
--- a/api/src/main/java/org/hawkular/inventory/api/filters/Related.java
+++ b/api/src/main/java/org/hawkular/inventory/api/filters/Related.java
@@ -31,7 +31,7 @@ public class Related<T extends Entity> extends Filter {
 
     private final T entity;
     private final String relationshipName;
-    private final String edgeId;
+    private final String relationshipId;
     private final EntityRole entityRole;
 
     /**
@@ -85,11 +85,11 @@ public class Related<T extends Entity> extends Filter {
      * Creates a filter for entities that are sources of at least one relationship with the specified id. The target
      * entity is not specified and can be anything.
      *
-     * @param edgeId the id of the edge
+     * @param relationshipId the id of the relationship
      * @return a new "related" filter instance
      */
-    public static Related<?> byEdgeWithId(String edgeId) {
-        return new Related<>(null, null, edgeId, EntityRole.SOURCE);
+    public static Related<?> byRelationshipWithId(String relationshipId) {
+        return new Related<>(null, null, relationshipId, EntityRole.SOURCE);
     }
 
     /**
@@ -139,10 +139,10 @@ public class Related<T extends Entity> extends Filter {
         return new Related<>(null, relationship.name(), EntityRole.TARGET);
     }
 
-    protected Related(T entity, String relationshipName, String edgeId, EntityRole entityRole) {
+    protected Related(T entity, String relationshipName, String relationshipId, EntityRole entityRole) {
         this.entity = entity;
         this.relationshipName = relationshipName;
-        this.edgeId = edgeId;
+        this.relationshipId = relationshipId;
         this.entityRole = entityRole;
     }
 
@@ -167,8 +167,8 @@ public class Related<T extends Entity> extends Filter {
     /**
      * @return the id of the relationship
      */
-    public String getEdgeId() {
-        return edgeId;
+    public String getRelationshipId() {
+        return relationshipId;
     }
 
     /**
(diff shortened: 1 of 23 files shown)


metlos:
Just a note for your consideration.

The current naive implementation of the `RelationshipService` requires fetching the start vertex upfront - I just hacked that up very quickly without much thought ;)

I.e. https://github.com/hawkular/hawkular-inventory/blob/future/impl-tinkerpop/src/main/java/org/hawkular/inventory/impl/tinkerpop/AbstractBrowser.java#L57

but consider this traversal:

```
inventory.tenants().get("tenant").relationships().named("kachny").environments().getAll().resources().getAll().entities();
```

in here, you first need to go to database to fetch the tenant and then continue with filtering until the `entities()` is called at which point you go to the db again.

Basically each call to `relationships()` requires a DB roundtrip in this naive impl.

Not to mention this query:

```
inventory.tenants().getAll().relationships().getAll().environments().getAll().entities();
```

in here there is even no single entity to feed the `RelationshipService` with.

I'd suggest trying to accomodate the approach from the rest of the `*Service` impls that just compose a list of filters that is then used to construct a gremlin query that does the whole of it at one go when trying to resolve the entities during `entity()`, `entities()` and other calls, possibly using a single access to the database (well, most possibly not, but tinkerpop could improve in future versions, couldn't they? ;) ).

api/src/main/java/org/hawkular/inventory/api/filters/Related.java
```diff
@@ -81,6 +82,17 @@
     }
 
     /**
+     * Creates a filter for entities that are sources of at least one relationship with the specified id. The target
+     * entity is not specified and can be anything.
+     *
+     * @param edgeId the id of the edge
+     * @return a new "related" filter instance
+     */
+    public static Related<?> byEdgeWithId(String edgeId) {
```
metlos: We have entities and relationships. Edge is an implementation detail in tinkerpop.
In another words, I'd call this `byRelationshipWithId(String id)`.

api/src/main/java/org/hawkular/inventory/api/filters/Related.java
```diff
@@ -148,6 +165,13 @@ public String getRelationshipName() {
     }
 
     /**
+     * @return the id of the relationship
+     */
+    public String getEdgeId() {
```
metlos: Again, IMHO, this would be better called `getRelationshipId`.

impl-tinkerpop/src/main/java/org/hawkular/inventory/impl/tinkerpop/RelationshipBrowser.java
```diff
@@ -0,0 +1,208 @@
...
+final class RelationshipBrowser<E extends Entity> extends AbstractBrowser<E> {
+
+    private RelationshipBrowser(Class<E> sourceClass, TransactionalGraph graph, FilterApplicator... path) {
+        super(graph, sourceClass, path);
+    }
+
+    public static Relationships.Single single(String id, HawkularPipeline<?, Vertex> source, Class<? extends Entity>
+            sourceClass, TransactionalGraph graph, FilterApplicator... path) {
+        if (null == id) {
+            throw new NullPointerException("unable to create Relationships.Single without the edge id.");
```
metlos: edge id -> relationship id

pilhuhn: Do not throw NPE here. Throw an IllegalArgumentException

api/src/main/java/org/hawkular/inventory/api/filters/Related.java
jkremser: done

impl-tinkerpop/src/main/java/org/hawkular/inventory/impl/tinkerpop/RelationshipBrowser.java
jkremser: done, done

metlos:
I am going to merge even if this is still the WIP. Please open a new PR with the rest of the changes once ready.

PR #31, opened 2015-03-06
HWKINVENT-15 - Relationship service #2

jkremser:
Follow-up pr

~~Don't merge yet.~~

### DIFF
--- a/api/src/main/java/org/hawkular/inventory/api/MultipleRelatableEntityBrowser.java
+++ b/api/src/main/java/org/hawkular/inventory/api/MultipleRelatableEntityBrowser.java
@@ -23,13 +23,21 @@ package org.hawkular.inventory.api;
  * @param <Entity> the type of the entity being browsed
  *
  * @author Lukas Krejci
+ * @author Jirka Kremser
  * @since 1.0
  */
 interface MultipleRelatableEntityBrowser<Entity> extends ResolvableToMany<Entity> {
 
     /**
-     * @return the (read) access interface to relationships of the entities on the current position in the inventory
-     * traversal.
+     * @return the (read) access interface to all (outgoing) relationships of the entities on the current position in
+     * the inventory traversal.
      */
     Relationships.Read relationships();
+
+    /**
+     * @param direction the direction of the relation (aka edge) This is needed because direction are not bidirectional.
+     * @return the (read) access interface to all relationships of the entities on the current position in
+     * the inventory traversal.
+     */
+    Relationships.Read relationships(Relationships.Direction direction);
 }
--- a/api/src/main/java/org/hawkular/inventory/api/MultipleRelatableEntityBrowser.java
+++ b/api/src/main/java/org/hawkular/inventory/api/MultipleRelatableEntityBrowser.java
@@ -35,7 +35,8 @@ interface MultipleRelatableEntityBrowser<Entity> extends ResolvableToMany<Entity
     Relationships.Read relationships();
 
     /**
-     * @param direction the direction of the relation (aka edge) This is needed because direction are not bidirectional.
+     * @param direction the direction of the relation (aka edge) This is needed because relationships are not
+     *                  bidirectional.
      * @return the (read) access interface to all relationships of the entities on the current position in
      * the inventory traversal.
      */
--- a/api/src/main/java/org/hawkular/inventory/api/Relationships.java
+++ b/api/src/main/java/org/hawkular/inventory/api/Relationships.java
@@ -60,19 +60,19 @@ public final class Relationships {
      */
     public enum Direction {
         /**
-         * Relative to the current position in the Hawkular pipeline, this value expresses such relationships
-         * that has me (the entity(ies) on the current pos) as a target(s).
+         * Relative to the current position in the inventory traversal, this value expresses such relationships
+         * that has me (the entity(ies) on the current pos) as a source(s).
          */
         outgoing,
 
         /**
-         * Relative to the current position in the Hawkular pipeline, this value expresses such relationships
-         * that has me (the entity(ies) on the current pos) as a source(s).
+         * Relative to the current position in the inventory traversal, this value expresses such relationships
+         * that has me (the entity(ies) on the current pos) as a target(s).
          */
         incoming,
 
         /**
-         * Relative to the current position in the Hawkular pipeline, this value expresses all the relationships
+         * Relative to the current position in the inventory traversal, this value expresses all the relationships
          * I (the entity(ies) on the current pos) have with other entity(ies).
          */
         both
--- a/api/src/main/java/org/hawkular/inventory/api/Relationships.java
+++ b/api/src/main/java/org/hawkular/inventory/api/Relationships.java
@@ -60,22 +60,20 @@ public final class Relationships {
      */
     public enum Direction {
         /**
-         * Expresses encapsulation of a set of entities in another entity.
-         * Used for example to express the relationship between a tenant and the set of its environments.
+         * Relative to the current position in the Hawkular pipeline, this value expresses such relationships
+         * that has me (the entity(ies) on the current pos) as a target(s).
          */
         outgoing,
 
         /**
-         * Expresses "instantiation" of some entity based on the definition provided by "source" entity.
-         * For example, there is a defines relationship between a metric definition and all metrics that
-         * conform to it.
+         * Relative to the current position in the Hawkular pipeline, this value expresses such relationships
+         * that has me (the entity(ies) on the current pos) as a source(s).
          */
         incoming,
 
         /**
-         * Expresses ownership. For example a resource owns a set of metrics, or a resource type owns a set
-         * of metric definitions. They do not contain it though, because more resources can own a single metric for
-         * example.
+         * Relative to the current position in the Hawkular pipeline, this value expresses all the relationships
+         * I (the entity(ies) on the current pos) have with other entity(ies).
          */
         both
     }
@@ -99,9 +97,7 @@ public final class Relationships {
     /**
      * Interface for accessing a single relationship in a writable manner
      */
-    public interface Single extends ResolvableToSingle<Relationship>,
-            BrowserBase<Tenants.ReadRelate, Environments.ReadRelate, Feeds.ReadRelate, MetricTypes.ReadRelate,
-                    Metrics.ReadRelate, Resources.ReadRelate, ResourceTypes.ReadRelate> {}
+    public interface Single extends ResolvableToSingle<Relationship> {}
 
     /**
      * Interface for traversing over a set of relationships.
@@ -117,8 +113,7 @@ public final class Relationships {
     /**
      * Provides read-write access to relationships.
      */
-    public interface ReadWrite extends ReadWriteRelationshipsInterface<Relationship, Relationship.Blueprint, Single,
-            Multiple> {
+    public interface ReadWrite extends ReadWriteRelationshipsInterface<Single, Multiple> {
         Multiple named(String name);
         Multiple named(WellKnown name);
     }
--- a/api/src/main/java/org/hawkular/inventory/api/Relationships.java
+++ b/api/src/main/java/org/hawkular/inventory/api/Relationships.java
@@ -54,6 +54,32 @@ public final class Relationships {
         owns
     }
 
+
+    /**
+     * The list of possible relationship (aka edges) direction. Relationships are not bidirectional.
+     */
+    public enum Direction {
+        /**
+         * Expresses encapsulation of a set of entities in another entity.
+         * Used for example to express the relationship between a tenant and the set of its environments.
+         */
+        outgoing,
+
+        /**
+         * Expresses "instantiation" of some entity based on the definition provided by "source" entity.
+         * For example, there is a defines relationship between a metric definition and all metrics that
+         * conform to it.
+         */
+        incoming,
+
+        /**
+         * Expresses ownership. For example a resource owns a set of metrics, or a resource type owns a set
+         * of metric definitions. They do not contain it though, because more resources can own a single metric for
+         * example.
+         */
+        both
+    }
+
     private interface BrowserBase<Tenants, Environments, Feeds, MetricTypes, Metrics, Resources, ResourceTypes> {
         Tenants tenants();
 
@@ -93,6 +119,7 @@ public final class Relationships {
      */
     public interface ReadWrite extends ReadWriteInterface<Relationship, Relationship.Blueprint, Single, Multiple> {
         Multiple named(String name);
+        Multiple named(WellKnown name);
     }
 
     /**
@@ -100,5 +127,6 @@ public final class Relationships {
      */
     public interface Read extends ReadInterface<Single, Multiple> {
         Multiple named(String name);
+        Multiple named(WellKnown name);
     }
 }
--- a/api/src/main/java/org/hawkular/inventory/api/Relationships.java
+++ b/api/src/main/java/org/hawkular/inventory/api/Relationships.java
@@ -117,7 +117,8 @@ public final class Relationships {
     /**
      * Provides read-write access to relationships.
      */
-    public interface ReadWrite extends ReadWriteInterface<Relationship, Relationship.Blueprint, Single, Multiple> {
+    public interface ReadWrite extends ReadWriteRelationshipsInterface<Relationship, Relationship.Blueprint, Single,
+            Multiple> {
         Multiple named(String name);
         Multiple named(WellKnown name);
     }
@@ -125,7 +126,7 @@ public final class Relationships {
     /**
      * Provides read access to relationships.
      */
-    public interface Read extends ReadInterface<Single, Multiple> {
+    public interface Read extends ReadRelationshipsInterface<Single, Multiple> {
         Multiple named(String name);
         Multiple named(WellKnown name);
     }
--- a/api/src/main/java/org/hawkular/inventory/api/SingleRelatableEntityBrowser.java
+++ b/api/src/main/java/org/hawkular/inventory/api/SingleRelatableEntityBrowser.java
@@ -23,8 +23,21 @@ package org.hawkular.inventory.api;
  * @param <Entity> the type of the entity being browsed
  *
  * @author Lukas Krejci
+ * @author Jirka Kremser
  * @since 1.0
  */
 interface SingleRelatableEntityBrowser<Entity> extends ResolvableToSingle<Entity> {
+
+    /**
+     * @return the (r/w) access interface to all (outgoing) relationships of the entities on the current position in
+     * the inventory traversal.
+     */
     Relationships.ReadWrite relationships();
+
+    /**
+     * @param direction the direction of the relation (aka edge) This is needed because direction are not bidirectional.
+     * @return the (r/w) access interface to all relationships of the entities on the current position in
+     * the inventory traversal.
+     */
+    Relationships.ReadWrite relationships(Relationships.Direction direction);
 }
--- a/api/src/main/java/org/hawkular/inventory/api/SingleRelatableEntityBrowser.java
+++ b/api/src/main/java/org/hawkular/inventory/api/SingleRelatableEntityBrowser.java
@@ -35,7 +35,8 @@ interface SingleRelatableEntityBrowser<Entity> extends ResolvableToSingle<Entity
     Relationships.ReadWrite relationships();
 
     /**
-     * @param direction the direction of the relation (aka edge) This is needed because direction are not bidirectional.
+     * @param direction the direction of the relation (aka edge) This is needed because relationships are not
+     *                  bidirectional.
      * @return the (r/w) access interface to all relationships of the entities on the current position in
      * the inventory traversal.
      */
--- a/api/src/main/java/org/hawkular/inventory/api/WriteRelationshipInterface.java
+++ b/api/src/main/java/org/hawkular/inventory/api/WriteRelationshipInterface.java
@@ -34,7 +34,7 @@ interface WriteRelationshipInterface<Single> {
      *
      * @param name the name of the relationship (label)
      * @param targetOrSource the the source/target entity (based on the chosen relationship direction) that the current
-     *                       entity (based on the hawkular pipeline) will be in the relationship with
+     *                       entity (based on the position in the inventory traversal) will be in the relationship with
      * @return access interface to the freshly created relationship
      *
      * @throws org.hawkular.inventory.api.RelationNotFoundException if the relationship already exists
@@ -46,7 +46,7 @@ interface WriteRelationshipInterface<Single> {
      *
      * @param name the well known name (Relationships.WellKnown) of the relationship
      * @param targetOrSource the the source/target entity (based on the chosen relationship direction) that the current
-     *                       entity (based on the hawkular pipeline) will be in the relationship with
+     *                       entity (based on the position in the inventory traversal) will be in the relationship with
      * @return access interface to the freshly created relationship
      *
      * @throws org.hawkular.inventory.api.RelationNotFoundException if the relationship already exists
--- a/api/src/main/java/org/hawkular/inventory/api/WriteRelationshipInterface.java
+++ b/api/src/main/java/org/hawkular/inventory/api/WriteRelationshipInterface.java
@@ -16,29 +16,42 @@
  */
 package org.hawkular.inventory.api;
 
+import org.hawkular.inventory.api.model.Entity;
 import org.hawkular.inventory.api.model.Relationship;
 
 /**
  * Generic methods to write access to relationships.
  *
- * @param <Entity> the entity type
- * @param <Blueprint> the blueprint type that supplies data necessary to create a new relationship
  * @param <Single> the access interface to a single relationship
  *
  * @author Jirka Kremser
  * @since 1.0
  */
-interface WriteRelationshipInterface<Entity, Blueprint, Single> {
+interface WriteRelationshipInterface<Single> {
 
     /**
      * Creates a new relationship at the current position in the inventory traversal.
      *
-     * @param blueprint the blueprint to be used to create the new relationship
+     * @param name the name of the relationship (label)
+     * @param targetOrSource the the source/target entity (based on the chosen relationship direction) that the current
+     *                       entity (based on the hawkular pipeline) will be in the relationship with
      * @return access interface to the freshly created relationship
      *
      * @throws org.hawkular.inventory.api.RelationNotFoundException if the relationship already exists
      */
-//    Single create(Blueprint blueprint) throws RelationNotFoundException;
+    Single linkWith(String name, Entity targetOrSource) throws RelationNotFoundException;
+
+    /**
+     * Creates a new relationship at the current position in the inventory traversal.
+     *
+     * @param name the well known name (Relationships.WellKnown) of the relationship
+     * @param targetOrSource the the source/target entity (based on the chosen relationship direction) that the current
+     *                       entity (based on the hawkular pipeline) will be in the relationship with
+     * @return access interface to the freshly created relationship
+     *
+     * @throws org.hawkular.inventory.api.RelationNotFoundException if the relationship already exists
+     */
+    Single linkWith(Relationships.WellKnown name, Entity targetOrSource) throws RelationNotFoundException;
 
     /**
      * Persists the provided relationship on the current position in the inventory traversal.
@@ -46,6 +59,8 @@ interface WriteRelationshipInterface<Entity, Blueprint, Single> {
      * @param relationship the relationship to update
      *
      * @throws org.hawkular.inventory.api.RelationNotFoundException if the relationship is not found in the database
+     * or if the source/target entity (based on the chosen relationship direction) doesn't correspond with the
+     * current position in the inventory traversal.
      */
     void update(Relationship relationship) throws RelationNotFoundException;
 
--- a/api/src/main/java/org/hawkular/inventory/api/WriteRelationshipInterface.java
+++ b/api/src/main/java/org/hawkular/inventory/api/WriteRelationshipInterface.java
@@ -1,0 +1,59 @@
+/*
+ * Copyright 2015 Red Hat, Inc. and/or its affiliates
+ * and other contributors as indicated by the @author tags.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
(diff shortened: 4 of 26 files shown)


api/src/main/java/org/hawkular/inventory/api/filters/RelationWith.java
```diff
@@ -0,0 +1,184 @@
...
+
+    public static Ids id(String id) {
+        return new Ids(id);
+    }
+
+    public static Ids ids(String... ids) {
+        return new Ids(ids);
+    }
+
+    public static Properties property(String property, String value) {
```
metlos, line 40: What about a filter on existence of a property regardless of its value?

api/src/main/java/org/hawkular/inventory/api/Relationships.java
```diff
@@ -54,6 +54,32 @@ private Relationships() {
         owns
     }
 
+
+    /**
+     * The list of possible relationship (aka edges) direction. Relationships are not bidirectional.
+     */
+    public enum Direction {
+        /**
```
metlos, line 62: Are the javadocs here correct? They seem to me a bit too concrete, because the meaning of a direction of a generic relationship is completely in the hands of the creator of the relationship.

api/src/main/java/org/hawkular/inventory/api/Relationships.java
```diff
@@ -91,14 +117,17 @@ private Relationships() {
     /**
      * Provides read-write access to relationships.
      */
-    public interface ReadWrite extends ReadWriteInterface<Relationship, Relationship.Blueprint, Single, Multiple> {
```
metlos, line 94: For the impossibility of commenting on unchanged lines, I comment here but really want to talk about the `Single` interface.

It provides access to single relationship and its "other ends". I am not sure the `*.ReadRelate` is correct interface for accessing the other ends, because the relationship is already established and a single relationship cannot have multiple sources or targets.

api/src/main/java/org/hawkular/inventory/api/WriteRelationshipInterface.java
```diff
@@ -0,0 +1,59 @@
...
+
+    /**
+     * Creates a new relationship at the current position in the inventory traversal.
+     *
+     * @param blueprint the blueprint to be used to create the new relationship
+     * @return access interface to the freshly created relationship
+     *
+     * @throws org.hawkular.inventory.api.RelationNotFoundException if the relationship already exists
+     */
+//    Single create(Blueprint blueprint) throws RelationNotFoundException;
```
metlos: Why is this commented out? Not implemented yet or do you envision another way of creating new relationships?

api/src/main/java/org/hawkular/inventory/api/filters/Filter.java
```diff
@@ -42,9 +42,10 @@
 public abstract class Filter {
     private static final Filter[] EMPTY = new Filter[0];
 
-    Filter() {
-
-    }
+//    we should not prohibit creating sub-classes of Filter in the implementations of this API
```
metlos: True, but on the other hand, this enables the user of the library to do the same and expect the implementation would work with it.

We could make that work by adding `boolean apply(Entity)` to the `Filter` class and implement an "in-memory" application of the filters. Implementations of the API can then choose to implement the specific filter classes in a more efficient manner and fall back to the `apply` method otherwise.

But I am not sure if that's really advisable solution. Our primary concern is the REST API which would lack such possibility.

Not sure how to approach this though. Your need for custom filters clear but at the same time I would not like to open the possibility for user-defined Filter impls. Maybe the `PathContext` and the accompanied classes would not use the `Filter`s directly but some kind of Filter decorator that would allow using the predefined and final set of filters from the API yet would allow for custom filters defined inside the implementation non-publicly? But that, too, is quite a big change.

api/src/main/java/org/hawkular/inventory/api/Relationships.java
jkremser, line 62: Right, good catch. It's copy&pasta 'typo'.

api/src/main/java/org/hawkular/inventory/api/filters/RelationWith.java
jkremser, line 40: Good idea, but let's not complicate it. This ought to be about the `Relationship service`, right? I am happy to add it afterwards.

api/src/main/java/org/hawkular/inventory/api/WriteRelationshipInterface.java
jkremser: IMO, the signature doesn't make sense. I commented it out because I wanted to discuss it before coding.

Imagine you navigate to some place in the graph, like

``` java
`Relationships.ReadWrite rw = inventory.tenants().get("foo").relationships();`
```

..and now you want to create a new relationship

``` java
Relationship.Blueprint rel = new Relationship.Blueprint("name", watSource, someTarget);
rw.create(rel);
```

It's strange, because I can specify arbitrary source in the blueprint and it should respect the path I am currently in. That was my thinking, so I suggest something like `link(String name, Entity target)` or perhaps some overloaded version also with the properties.

metlos: Yeah, the blueprint is incorrect ;) It should not specify the source because that is deduced from the path as you correctly note.

api/src/main/java/org/hawkular/inventory/api/filters/Filter.java
jkremser: I needed to add the `JumpInOutFilter` for "jumping" to the edges or from the edges during the graph traversal. At the same time the filter should be an implementation detail in the library implementation (default visibility).

> True, but on the other hand, this enables the user of the library to do the same and expect the implementation would work with it.

JDoc? Right now the `FilterApplicator` kind of serves as a white list of allowed filters. If the library consumer naively extends a filter with the hope it'll work. It fails with the first usage of the filter with

`IllegalArgumentException("Unsupported filter type " + filter.getClass());`

I'd argue that using the library != extending it's classes and pass this derived classes naively to the library and expects it'll work. It's not a big deal if we comment it in JDoc, I think.

The "in-memory" application sound interesting. Do you mean applying the filters on the network of interconnected Java objects or using something like ispn?

api/src/main/java/org/hawkular/inventory/api/Relationships.java
jkremser, line 94: I see, I agree it is quite weird. It's actually similar for the `Multiple` iface, but it makes more sense there as the .tenants()/.environments()/.resources()/.etcs() methods works as an additional filter. Basically the same way as the `TargetOfType` filter now works (provided the outgoing relationships were chosen).

For the Single it's the same, it's just an additional filter on a type, I admit, it doesn't make much sense for the Single, perhaps something like `getMeTheOtherEndOfThisRelationshipPlease()` makes more sense here. With the cost of losing the type safety, or perhaps this would be solvable as well with some lookahead to the "target"'s type in the hawk-pipe.

api/src/main/java/org/hawkular/inventory/api/Relationships.java
jkremser, line 62: Actually, I don't like the discrepancy between the "source/target/both" and "incoming/outgoing/both" nomenclature, but I wasn't able to come up with a better meaningful names for the `.relationships(<<value>>)` "sourced" and "targeted" sounds quite confusing to me.

wdyt? What about using the "sourc(ed)/target(ed)/both" when speaking about entities and using the Related filter and "incoming/outgoing/both" when talking about the relationships itself?

api/src/main/java/org/hawkular/inventory/api/Relationships.java
metlos, line 94: The `getMeTheOtherEndOfThisRelationshipPlease()` method could be simply replaced by `...entity().getTarget())`.

So maybe for relationships the `Single` interface actually just degrades to `ResolvableToSingle`

api/src/main/java/org/hawkular/inventory/api/filters/Filter.java
metlos: I guess Javadoc is the best solution here ;)

The in-memory application is basically the same idea as `GremlinPipeline.filter()` method. It just applies a predicate to the intermediate results of the pipeline.

api/src/main/java/org/hawkular/inventory/api/Relationships.java
metlos, line 62: For me, `incoming`, `outgoing` and `both` is clear enough. Some entity is a target of a relationship outgoing from this entity which is a source of the relationship.

Or conversely, some entity is a source of a relationship that is incoming to this entity, which is the target of the relationship.

I.e. source/target express the "position" of the entities, while incoming/outgoing express the direction of the relationship.

api/src/main/java/org/hawkular/inventory/api/Relationships.java
jkremser, line 94: right, or `...relationships(Relationships.Direction.incoming).entity().getSource()`

api/src/main/java/org/hawkular/inventory/api/WriteRelationshipInterface.java
jkremser: I've added `Single linkWith(String name, Entity targetOrSource)` and `Single linkWith(Relationships.WellKnown name, Entity targetOrSource)`

Properties can be changed by modifying the `Relationship` instance that is returned when calling the `linkWith(..).entity()` methods and calling the `update`. The `update()` is not yet done.

metlos: K, so the Relationship.Blueprint is not needed anymore, am I right?

jkremser: Yes, it's not needed. I am working on the update method and it looks like it's not possible to update the relationship http://stackoverflow.com/a/18425028/1594980. It is possible to delete it and create a new one, but it gets different id and it's questionable if it should be called updating if the identity changes.

What is possible is to change the property map, so that's what I am doing.. +all the checks

api/src/main/java/org/hawkular/inventory/api/Relationships.java
```diff
@@ -54,6 +54,30 @@ private Relationships() {
         owns
     }
 
+
+    /**
+     * The list of possible relationship (aka edges) direction. Relationships are not bidirectional.
+     */
+    public enum Direction {
+        /**
+         * Relative to the current position in the Hawkular pipeline, this value expresses such relationships
```
metlos: Hawkular pipeline is an impl detail in the tinkerpop impl. I'd not mention it here in the API. Maybe use "inventory traversal" instead?

api/src/main/java/org/hawkular/inventory/api/Relationships.java
```diff
@@ -54,6 +54,30 @@ private Relationships() {
...
+         * that has me (the entity(ies) on the current pos) as a target(s).
+         */
+        outgoing,
+
+        /**
+         * Relative to the current position in the Hawkular pipeline, this value expresses such relationships
```
metlos: same as above

api/src/main/java/org/hawkular/inventory/api/Relationships.java
```diff
@@ -54,6 +54,30 @@ private Relationships() {
...
+         * that has me (the entity(ies) on the current pos) as a source(s).
+         */
+        incoming,
+
+        /**
+         * Relative to the current position in the Hawkular pipeline, this value expresses all the relationships
```
metlos: same as above

api/src/main/java/org/hawkular/inventory/api/MultipleRelatableEntityBrowser.java
```diff
@@ -23,13 +23,21 @@
...
-     * @return the (read) access interface to relationships of the entities on the current position in the inventory
-     * traversal.
+     * @return the (read) access interface to all (outgoing) relationships of the entities on the current position in
+     * the inventory traversal.
      */
     Relationships.Read relationships();
-}
+
+    /**
+     * @param direction the direction of the relation (aka edge) This is needed because direction are not bidirectional.
```
metlos: s/because direction/because relationships/

api/src/main/java/org/hawkular/inventory/api/Relationships.java
```diff
@@ -54,6 +54,30 @@ private Relationships() {
...
     }
 
+
+    /**
+     * The list of possible relationship (aka edges) direction. Relationships are not bidirectional.
+     */
+    public enum Direction {
+        /**
+         * Relative to the current position in the Hawkular pipeline, this value expresses such relationships
+         * that has me (the entity(ies) on the current pos) as a target(s).
```
metlos: Isn't this actually the other way round?
I was assuming that `...relationships(outgoing).named(defines).resources().entities()` would get me all the resources that are _targets_ of defines relationships going out from "me" - i.e. "me" is the source of the relationships.

api/src/main/java/org/hawkular/inventory/api/SingleRelatableEntityBrowser.java
```diff
@@ -23,8 +23,21 @@
...
 interface SingleRelatableEntityBrowser<Entity> extends ResolvableToSingle<Entity> {
+
+    /**
+     * @return the (r/w) access interface to all (outgoing) relationships of the entities on the current position in
+     * the inventory traversal.
+     */
     Relationships.ReadWrite relationships();
+
+    /**
+     * @param direction the direction of the relation (aka edge) This is needed because direction are not bidirectional.
```
metlos: copy&paste from Multiple...

api/src/main/java/org/hawkular/inventory/api/filters/Filter.java
```diff
@@ -34,7 +34,15 @@
...
  *
  * <p>The implementations of the Hawkular inventory API are supposed to support filtering by {@link Related},
- * {@link With.Ids} and {@link With.Types}.
+ * {@link With.Ids} and {@link With.Types}. There is also a sub-class of filters for the relation filtering {@link
+ * RelationFilter}.
+ *
+ * To create these filters, feel free to use the static helper methods defined on {@link With}.
+ * <p>
+ * Note: Additional information for the library consumers.
+ * Don't extend this class with the hope the new filter will work without any further work, It needs to be registered
```
metlos: This leaves hope that after "registration" it might work. I'd use something stronger as:
"Don't extend this class with hope that the new filter will work. This class is extendable only for the benefit of the API implementations that can reuse it internally. For the users of the API, only the subclasses of Filter declared directly in the API are available."

api/src/main/java/org/hawkular/inventory/api/filters/RelationFilter.java
```diff
@@ -0,0 +1,77 @@
...
+ * {@link org.hawkular.inventory.api.filters.RelationWith.Properties},
+ * {@link org.hawkular.inventory.api.filters.RelationWith.Ids},
+ * {@link org.hawkular.inventory.api.filters.RelationWith.TargetOfType},
+ * {@link org.hawkular.inventory.api.filters.RelationWith.SourceOfType} and
+ * {@link org.hawkular.inventory.api.filters.RelationWith.SourceOrTargetOfType}.
+ *
+ * To create these filters, feel free to use the static helper methods defined on {@link RelationWith}.
+ * <p>
+ * Note: Additional information for the library consumers.
+ * Don't extend this class with the hope the new filter will work without any further work, It needs to be registered
```
metlos: The same note as with `Filter`.

api/src/main/java/org/hawkular/inventory/api/Relationships.java
jkremser: correct, I missed that

impl-tinkerpop/src/main/java/org/hawkular/inventory/impl/tinkerpop/FilterApplicator.java
```diff
@@ -47,8 +49,10 @@ private FilterApplicator(Type type) {
     }
 
     public static FilterApplicator with(Type type, Filter filter) {
+        // note: this is a static method so this can't be achieved elegantly with polymorphism
```
metlos: How would you replace the below if-chain with polymorphic calls? I don't like this solution either, so any improvement is welcome :)

impl-tinkerpop/src/main/java/org/hawkular/inventory/impl/tinkerpop/RelationshipBrowser.java
```diff
@@ -50,158 +53,130 @@
...
+                HawkularPipeline<?, Edge> edges = b.source().cast(Edge.class);
 
-                HawkularPipeline<?, Edge> edges = null == named ? b.source().bothE() : b.source().bothE(named);
                 Stream<Relationship> relationshipStream = StreamSupport
                         .stream(edges.spliterator(), false)
-                        .map(edge -> new Relationship(edge.getId().toString(), edge.getLabel(),
-                                convert(edge.getVertex(Direction.OUT)), convert(edge.getVertex(Direction.IN))));
+                        .map(edge -> {
+                            Relationship relationship = new Relationship(edge.getId().toString(), edge.getLabel(),
+                                    convert(edge.getVertex(Direction.OUT)), convert(edge.getVertex(Direction.IN)));
```
metlos, line 104: IMHO, this is the other way around. `Edge.getVertex(OUT)` returns the "tail" of the edge according to the javadoc, so in Hawkular-speak, I'd say that that is the target of the relationship.

impl-tinkerpop/src/main/java/org/hawkular/inventory/impl/tinkerpop/FilterVisitor.java
```diff
@@ -109,4 +112,100 @@ public void visit(HawkularPipeline<?, ?> query, With.Types types) {
...
+
+        q2.or(typeChecks).back(label);
+    }
+
+    public void visit(HawkularPipeline<?, ?> query, RelationshipBrowser.JumpInOutFilter filter) {
+        final boolean jumpFromEdge = filter.isFromEdge();
+        switch (filter.getDirection()) {
+            case incoming:
+                if (jumpFromEdge) {
+                    query.outV();
```
metlos, line 190: Why is an incoming direction of a relationship modelled as jumping to the "out" vertex of an edge?

impl-tinkerpop/src/main/java/org/hawkular/inventory/impl/tinkerpop/RelationshipBrowser.java
```diff
@@ -50,158 +53,130 @@
...
-                                .class, ResourcesService.class);
-            }
-
-            @Override
-            public ResourceTypes.ReadRelate resourceTypes() {
-                return (ResourceTypes.ReadRelate)b.<ResourceTypesService>getService(EdgeFilter.ID, relationshipId,
-                        ResourceType.class,
-                        ResourceTypesService.class);
+                Relationship relationship = new Relationship(edge.getId().toString(), edge.getLabel(), convert(edge
+                         .getVertex(Direction.OUT)), convert(edge.getVertex(Direction.IN)));
```
metlos, line 78: I think this is the other way around... `Edge.getVertex(OUT)` fetches the vertex on the "tail" end of the edge, which IMHO translates to the target of the relationship in the Hawkular-speak.

impl-tinkerpop/src/main/java/org/hawkular/inventory/impl/tinkerpop/RelationshipBrowser.java
jkremser, line 104: This looks good to me. 

assuming:
source -----e----->  target

`e.getVertex(Direction.OUT)` would be `source`
`e.getVertex(Direction.IN)` would be `target`

The Relationship constructors takes `(String id, String name, Entity source, Entity target)`

impl-tinkerpop/src/main/java/org/hawkular/inventory/impl/tinkerpop/RelationshipService.java
```diff
@@ -16,86 +16,219 @@
...
+        }
+
+        // check if the source/target vertex of the relationship is on the current position in hawk-pipe. If not use the
+        // `ifThenElse` for returning an empty iterator and fail subsequent querying
+        PipeFunction<Vertex, Boolean> ifFunction = vertex -> {
+            String uid = vertex.getProperty(Constants.Property.uid.name());
+            boolean sourceOk = uid.equals(relationship.getSource().getId());
+            boolean targetOk = uid.equals(relationship.getTarget().getId());
+            boolean retBool = direction == outgoing ? sourceOk : direction ==
+                    incoming ? targetOk : (sourceOk || targetOk);
```
metlos, line 173: shouldn't this be `sourceOk && targetOk` in case of `both` direction?

impl-tinkerpop/src/main/java/org/hawkular/inventory/impl/tinkerpop/RelationshipService.java
```diff
@@ -16,86 +16,219 @@
...
+            return retBool;
+        };
+        PipeFunction<Vertex, ?> thenFunction = v -> v;
+        PipeFunction<Vertex, ?> elseFunction = v -> Collections.<Vertex>emptyList().iterator();
+        source().ifThenElse(ifFunction, thenFunction, elseFunction);
```
metlos: Why are you not assigning this to any variable? `source()` creates a new pipeline everytime...

impl-tinkerpop/src/main/java/org/hawkular/inventory/impl/tinkerpop/RelationshipService.java
metlos, line 173: scratch that :)

impl-tinkerpop/src/test/java/org/hawkular/inventory/impl/tinkerpop/test/BasicTest.java
```diff
@@ -283,8 +295,12 @@ public void testRelationshipServiceGet() throws Exception {
             "going from  environments must have uid equal to 'test'. Was: " + contains.getTarget().getId();
         assert "contains" .equals(contains.getName()) : "Name of the relation must be 'contains'.";
 
-        contains = inventory.tenants().getAll().environments().get("test").relationships().get("13").entity();
-        assert contains == null : "There should not be any edge with id 13.";
+        try {
+            inventory.tenants().getAll().environments().get("test").relationships().get("13").entity();
```
metlos, line 299: How did you come up with 13?

impl-tinkerpop/src/test/java/org/hawkular/inventory/impl/tinkerpop/test/BasicTest.java
```diff
@@ -318,16 +334,149 @@ public void testRelationshipServiceNamed2() throws Exception {
...
+                .get("playroom2").metrics().get("playroom2_size").relationships(Relationships.Direction.outgoing)
+                .named("yourMom").entities().iterator().next();
+        assert rel1.getId().equals(rel3.getId()) && someValue.equals(rel3.getProperties().get(someKey))
+                : "There should be the property with key 'k3y' and value 'v4lu3'";
 
+        try {
+            inventory.tenants().get("com.example.tenant").environments().get("test").resources()
+                    .get("playroom2").metrics().get("playroom1_size").relationships(Relationships.Direction.both)
+                    .update(rel1);
+            assert !!!true : "It shouldn't be possible to update an edge that is not on the current position in the " +
```
metlos, line 407: Funny ;) Also I don't fully understand why this doesn't fail in light of my comment at the `source().ifThenElse()` in `update()`.

impl-tinkerpop/src/main/java/org/hawkular/inventory/impl/tinkerpop/RelationshipService.java
jkremser: Where exactly? In the switch?

jkremser: Oh, I see.. right, L178 - st call and  the switch - 2nd call

jkremser: btw. the `RelationshipService.linkWith()` needs to go twice to the "db"
1st time: `convert(targetOrSource);` on L118
2nd: the switch in the query 

It's because of the signature of those `link{Out|In|Both}` methods that expects a vertex as a parameter, so I need to look it up first. Perhaps this could be optimized using those overloaded variants with the named step as a parameter. But let's keep it simple and iterate over it, wdyt?

impl-tinkerpop/src/main/java/org/hawkular/inventory/impl/tinkerpop/RelationshipBrowser.java
metlos, line 78: Ah, it is me who misunderstood blueprints' javadoc. I thought "tail" would mean the "end" of the edge. Their class-level javadoc explains it better.

impl-tinkerpop/src/test/java/org/hawkular/inventory/impl/tinkerpop/test/BasicTest.java
jkremser, line 299: 13 is the magic number ;) It's mentioned in the jdoc, 

//here we assume the edge IDs in advance to be 10 and 14 and there is not edge with id 13

The underlying graph creates the IDs deterministically and these numbers are just assumed for the graph that is set up in the `setupData` method. Not super smart, I admit. Perhaps better strategy would be to fetch some relationship by some other criteria, find out the id and then fetch it by id and compare the two.

impl-tinkerpop/src/main/java/org/hawkular/inventory/impl/tinkerpop/RelationshipBrowser.java
metlos, line 104: :+1:

impl-tinkerpop/src/test/java/org/hawkular/inventory/impl/tinkerpop/test/BasicTest.java
jkremser, line 299: btw. I found much better to test some general properties of the graph than some implementation details. What about using something for checking those properties on randomly generated graphs? ScalaCheck test framework has so called generators, I think it's inspired by QuickCheck, that are able to generate random data structures (lists/trees/heaps/graphs..). 

And tests verifies some general properties like:
- when adding 1 edge to a random graph, resulting graph has number of edges equal to the original graph + 1,
- when removing a tenant, all it's data are deleted (no leaking, we don't support it yet) and similar.

Here is something http://www.scala-graph.org/guides/test.html but it can be easily written for our needs.

impl-tinkerpop/src/main/java/org/hawkular/inventory/impl/tinkerpop/FilterApplicator.java
jkremser: Oh I missed the fact that `Filter` is in the "API" and `FilterApplicator` is in the "API Impl". I don't know, what about something like this:

http://pastebin.test.redhat.com/268206

Far from perfect, though. It has nothing to do with polymorphism, but it's good to have all the binding at one place. Perhaps the visit method could be implemented in those applicator sub-classes as well.

impl-tinkerpop/src/main/java/org/hawkular/inventory/impl/tinkerpop/RelationshipService.java
metlos: > btw. the `RelationshipService.linkWith()` needs to go twice to the "db"
> 1st time: `convert(targetOrSource);` on L118
> 2nd: the switch in the query
> 
> It's because of the signature of those `link{Out|In|Both}` methods that
> expects a vertex as a parameter, so I need to look it up first. Perhaps
> this could be optimized using those overloaded variants with the named step
> as a parameter. But let's keep it simple and iterate over it, wdyt?

+1, let's go with this now and see how it performs.

impl-tinkerpop/src/main/java/org/hawkular/inventory/impl/tinkerpop/FilterVisitor.java
jkremser, line 190: These `JumpInOutFilter` filters should be used for "changing the context" of what will be subsequently narrowed by other `has('property','value')` filters. By changing the context I mean whether it's an edge or a vertex. It's like a half-step in the inventory traversal.

I was lazy to create 2 separate filters, one for going from the vertex to the edges and second for going from edge to it's adjacent vertices.

Imagine following calls:
`...relationships(incoming).getAll(f1,f2,f3).tenants();`

this, vaguely, means:
1. jump to incoming edges [this must be done correctly]
2. perform filters f1,f2 and f3
3. jump from incoming edges (to vertices) [this must be done correctly]
4. filter vertices by type (must be tenants)

So for 'incoming' type of edge there are two cases:
1) jumping to edges `query.inE();` get me all edges that are incomming to the vertex on the current position
2) jumping from the edges `query.outV();` get me all the vertices from which these edges starts (source entities)

impl-tinkerpop/src/test/java/org/hawkular/inventory/impl/tinkerpop/test/BasicTest.java
jkremser, line 407: The test is passing because it throws the `RelationNotFoundException` exception, but from a different reason than the source or target mismatch. It fails because of the wrong rel.id. To be more precise, because of non-existent relation on the current position with such an id.

I'll add another test testing the scenario with the target/source vertex mismatch. You're right that it wasn't working because of calling the `source()` twice, because the `ifThenElse` pipe had no effect, when starting the pipe after few lines from scratch. Good catch!

jkremser:
50th comment! :trollface:

impl-tinkerpop/src/test/java/org/hawkular/inventory/impl/tinkerpop/test/BasicTest.java
metlos, line 299: +1, I also think we should really move the testsuite to the API and make it "abstract" in that it would need to be "seeded" with a concrete implementation.

The tinkerpop impl would then just instantiate the testsuite, seed it with the tinkerpop inventory impl and run it.

This way we define the required behavior generically and impls will have something to conform to.

Blueprints testsuite is written this way btw.

PR #33, opened 2015-03-06
Readme for the future branch

metlos:


pilhuhn:
Why another chat system? Do we really need one chat per component? Individual repo?

metlos:
No, not really. Jirka added it for people having trouble with IRC, but then there's webchat.freenode.net, so maybe gitter.im is kinda redundant.

jkremser:
We can remove it, if noone is using it in the future. The more badges on the Readme page, the cooler project looks like ;)

PR #34, opened 2015-03-13
Update/Delete on entities

metlos:


impl-tinkerpop/src/main/java/org/hawkular/inventory/impl/tinkerpop/AbstractSourcedGraphService.java
```diff
@@ -78,6 +93,121 @@ public Single create(Blueprint blueprint) {
...
+        Constants.Type type = Constants.Type.of(entity);
+        List<String> mappedProperties = Arrays.asList(type.getMappedProperties());
+
+        entity.getProperties().keySet().forEach(k -> {
+            if (mappedProperties.contains(k)) {
+                throw new IllegalArgumentException("Property '" + k + "' is reserved. Cannot set it to a custom value");
+            }
+        });
+
+        Vertex vertex = convert(entity);
```
jkremser, line 102: `convert(entity)` might an expensive op (assuming we don't have any global index) what about using 

``` java
        HawkularPipeline<?, Vertex> vertices = source().hasUid(entity.getId());
        if (!vertices.hasNext()) throw ENFE..
        Vertex vertex = vertices.next();
```

impl-tinkerpop/src/main/java/org/hawkular/inventory/impl/tinkerpop/AbstractSourcedGraphService.java
```diff
@@ -78,6 +93,121 @@ public Single create(Blueprint blueprint) {
...
+        if (vertex == null) {
+            throw new EntityNotFoundException(entity.getClass(), FilterApplicator.filters(pathContext.path));
+        }
+
+        Set<String> toRemove = vertex.getPropertyKeys();
+        toRemove.removeAll(entity.getProperties().keySet());
+
+        toRemove.forEach(vertex::removeProperty);
+        entity.getProperties().forEach(vertex::setProperty);
```
jkremser, line 111: `ElementHelper.setProperties(Element element, Map<String, Object> properties)` ?

impl-tinkerpop/src/main/java/org/hawkular/inventory/impl/tinkerpop/AbstractSourcedGraphService.java
```diff
@@ -104,4 +234,90 @@ protected void removeRelationship(Constants.Type typeInSource, Relationships.Wel
...
+            Iterator<Vertex> queueToProcess = peekNonEmpty();
+            if (queueToProcess == null) {
+                return false;
+            }
+
+            Vertex toProcess = queueToProcess.next();
+
+            pending.add(toProcess.getVertices(direction, edgeLabels).iterator());
+
+            action.accept(toProcess);
```
jkremser: Nice! I am not sure if it'll work also with Titan, though. I think the vertices are lazily fetched when needed. It works with our current impl, because it's just a net of objects.

btw. I did the same (assuming the getVertices().getVertices().getVertices() will work) somewhere in the `RelationshipService`, but I am not convinced it'll work with C*.

Have you considered using some Gremlin-fu for the transitive closure? It's Groovy, not Java.

``` Groovy
g = new TinkerGraph()
v1 = g.addVertex([name: 'one'])
v2 = g.addVertex([name: 'two'])
v3 = g.addVertex([name: 'three'])
v4 = g.addVertex([name: 'four'])
v5 = g.addVertex([name: 'five'])

v1.addEdge('contains', v2)
v2.addEdge('contains', v3)
v3.addEdge('contains', v4)
v4.addEdge('contains', v5)

v1.out('contains').loop(1){true}{true}.name
==>two
==>three
==>four
==>five
```

⋅⋅⋅ It's not smart enough to detect cycles, but this can be handled with some iteration limit `{it.loops < 10}{true}`

jkremser: Also removing the vertices directly in Gremlin should be more efficient than doing it 1by1 with the need of their looking up (global index issue)

st. like
`v1.out('contains').loop(1){true}{true}.remove()`
`v1.remove()`

metlos: I just tried and it's not working on Titan (and because my Stream uses the same approach I assume it wouldn't work, too).

In your example graph above, running:

``` java
v1.out("contains").loop(1){true}{true}.remove()
```

will barf with "unable to access a removed vertex" error. That is because the loop (as well as my stream) will return vertices as soon as it finds them - and the pipe will call `remove()` on them as they flow in. So you first remove a vertex and then want to find other vertices by looking at the "contains" edges of the element you just removed.

This needs to go "bottom up" instead. Any ideas for some Gremlin magic to do that?

jkremser: Right, I've just tried on `titan-server-0.4.4` and it removed only v2. Same holds for `v1.out('contains').loop(1){true}{true}.each{g.removeVertex(it)}`

This works for me 

``` groovy
v1.out('contains').loop(1){it.loops < 10}{true}.order{-it.a.path.toList().unique{a, b -> a <=> b}.size <=> it.b.path.toList().unique{a, b -> a <=> b}.size}.remove()
v1.remove()
```

It sorts the vertices by the path from v1 (vertices with the longest path go first). So basically it removes everything starting from the leaves.

metlos: I managed to implement a DFS search that only holds a subset of the vertices and unresolved queries for vertices in memory (to be pushed shortly).

I wonder what is more efficient. If that custom DFS or your example that needs to load all the vertices to memory.

impl-tinkerpop/src/main/java/org/hawkular/inventory/impl/tinkerpop/AbstractSourcedGraphService.java
metlos, line 102: Ah, forgot about this one... will do something about it...

metlos, line 102: Well, the example you gave is pretty much exactly what `convert(Entity)` does so I think we're cool here.

Btw. IMHO it is always needed to use both `type` and `uid` constraints on `source()` because the `uid` is user provided and I can easily imagine two entities of different types having the same uid. Of course, `source()` puts context (tenant, env, ...) into the search so the probability of clash is lowered by that but still.

metlos:
I think this is ready to merge. Do you have any other suggestions?

jkremser:
> I think this is ready to merge. Do you have any other suggestions?
> :shipit:

PR #43, opened 2015-04-03
[HWKINVENT-26] Bus Integration for Inventory

metlos:
Including wiring it up in the REST war.

For this to work in Kettle, we'll need to update its config to create the new message topics.


metlos:
Oh, this builds on top of #42, so the number of commits is higher than it actually was.

Bus integration is done in f9e1793 and 2e544a8.

PR #47, opened 2015-04-09
When creating a resource with some additional metadata stored in the pro...

jkremser:
...perties hash, it was ignored.


metlos:
Huh, did the very same thing, only for all entity types in #48 ;)

jkremser:
ouch, me 2 :]

jkremser:
lol, it's almost the same

metlos:
After discussing with Jirka, I'm closing this in favor of #48. Thanks for the effort and confirming the approach ;)

PR #53, opened 2015-04-16
HWKINVENT-30 - some REST improvements

jkremser:
retrieving a single tenant and retrieving the resources of non-existent resource type now works properly

### DIFF
--- a/rest-servlet/src/main/java/org/hawkular/inventory/rest/RestResourceTypes.java
+++ b/rest-servlet/src/main/java/org/hawkular/inventory/rest/RestResourceTypes.java
@@ -23,6 +23,7 @@ import com.wordnik.swagger.annotations.ApiParam;
 import com.wordnik.swagger.annotations.ApiResponse;
 import com.wordnik.swagger.annotations.ApiResponses;
 import org.hawkular.inventory.api.Inventory;
+import org.hawkular.inventory.api.ResourceTypes;
 import org.hawkular.inventory.api.model.MetricType;
 import org.hawkular.inventory.api.model.Resource;
 import org.hawkular.inventory.api.model.ResourceType;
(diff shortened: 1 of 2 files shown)


rest-servlet/src/main/java/org/hawkular/inventory/rest/RestResourceTypes.java
```diff
@@ -112,8 +113,9 @@ public ResourceType get(@PathParam("tenantId") String tenantId,
     public Set<Resource> getResources(@PathParam("tenantId") String tenantId,
                                       @PathParam("resourceTypeId") String resourceTypeId) {
 
-        return inventory.tenants().get(tenantId).resourceTypes().get(resourceTypeId)
-                .resources().getAll().entities();
+        ResourceTypes.Single single = inventory.tenants().get(tenantId).resourceTypes().get(resourceTypeId);
```
metlos, line 116: This is a more general problem, IMHO, with the way we resolve `entities()` method in the tinkerpop impl. I am not sure we should resolve to one-off fixups like this, because the same issue is still present for all other entity types.

jkremser, line 116: agreed, but I'd differentiate the java api and rest api. I think it's perfectly fine to return the empty set if the filters in the method chain (pretty arbitrary and general approach) doesn't match any entities. 

On the other hand, IMO, we can't do the same for REST endpoints that should return "all resources of given resource type". Returning HTTP 200 with empty JSON array for non-existent resource types is just wrong. (currently what is happening is this: GET http://foo/bar/baz = 404, but http://foo/bar/baz/mazz = 200) Our rest api is just coarse grained, we don't allow for any combination of filters as we do for the Java API. I feel the difference here.

metlos, line 116: k, you persuaded me :smile:  But in that case we should do the same thing for all other entity types...

http://.../inventory/no-tenant/environments should also return 404, etc.

jkremser, line 116: ok, now I see your point. I thought, you were talking about other entity types in the Java API. 

It's pretty much everywhere in the REST:
/inventory/no-tenant/test/{resources|metrics}
/inventory/test/no-env/{resources|metrics}
/inventory/no-tenant/no-env/resources/no-resource/metrics
/inventory/no-tenant/metricTypes
/inventory/no-tenant/resourceTypes/no-resType/metricTypes
...
it all returns [] 200 :/

metlos:
So this is still only done for resources and not for the rest of the entity types?

PR #54, opened 2015-04-20
[HWKINVENT-31] Metrics and resources can now live under feeds, too.

metlos:
This adds the possibility for the resources and metrics to be contained
within feeds, too.

This adds the ability to have "feed-local" resource and metric ids
without any special ID requirements. I.e. two feeds can declare
resources with the same ID easily as long as the resources are contained
within the respective feeds in inventory.

The ability to have feedless resources and metrics has been preserved to
enable ad-hoc user-defined resources.

This is the first commit with a draft implementation of the
functionality.

Missing pieces:
- [x] unit tests
- [x] REST API updates
- [x] Javadocs


metlos:
Commit 6b8c1f5 is kinda big and added the ability to fetch all resources or metrics directly from the environment regardless of whether the they contained directly to the environment or are contained in one of the feeds of the environment.

This is IMHO important, because people will want to query across all entities in an environment and will not want to do a number of feed-bound queries for that.

But the commit is big because it required quite a bit of change in the tinkerpop impl to keep supporting the lazy construction of queries.

PR #55, opened 2015-04-21
Exposing relationships via REST api

jkremser:
not finished


rest-servlet/pom.xml
```diff
@@ -110,8 +110,14 @@
       <scope>provided</scope>
     </dependency>
 
-    <!-- docs -->
+    <dependency>
+      <groupId>com.github.jsonld-java</groupId>
+      <artifactId>jsonld-java</artifactId>
+      <!--<version>0.5.1</version> -->
+      <version>0.6.0-SNAPSHOT</version>
+    </dependency>
```
jkremser: this will probably go away, it's not needed

rest-servlet/src/main/java/org/hawkular/inventory/rest/RestMetricTypes.java
```diff
@@ -132,4 +137,34 @@ public Response delete(@PathParam("tenantId") String tenantId,
...
+
+    @GET
+    @Path("/{tenantId}/metricTypes/{metricTypeId}/relationships")
+    @ApiOperation("Retrieves all relationships of given metric type.")
+    @ApiResponses({
+            @ApiResponse(code = 200, message = "OK"),
+            @ApiResponse(code = 404, message = "Tenant or metric type not found", response = ApiError.class),
+            @ApiResponse(code = 500, message = "Server error", response = ApiError.class)
+    })
+    public Response getEnvironmentRelations(@PathParam("tenantId") String tenantId,
```
metlos: copy&paste - method name, resourceTypeId..

rest-servlet/src/main/java/org/hawkular/inventory/rest/RestMetrics.java
```diff
@@ -152,4 +157,38 @@ public Response deleteMetric(@PathParam("tenantId") String tenantId,
...
+    @GET
+    @Path("/{tenantId}/{environmentId}/resources/{resourceId}/relationships")
+    @ApiOperation("Retrieves all relationships of given metric.")
+    @ApiResponses({
+            @ApiResponse(code = 200, message = "OK"),
+            @ApiResponse(code = 404, message = "Tenant, environment or metric doesn't exist",
+                    response = ApiError.class),
+            @ApiResponse(code = 500, message = "Server error", response = ApiError.class)
+    })
+    public Response getEnvironmentRelations(@PathParam("tenantId") String tenantId,
```
metlos: copy&paste

rest-servlet/src/main/java/org/hawkular/inventory/rest/RestRelationships.java
```diff
@@ -0,0 +1,150 @@
...
+public class RestRelationships {
+
+    @Inject @ForRest
+    private Inventory inventory;
+
+    @GET
+    @Path("/")
+    @ApiOperation("Lists all relationships")
+    @ApiResponses({
+            @ApiResponse(code = 200, message = "The list of tenants"),
```
metlos: This is not limited to just tenants, am I right?

rest-servlet/src/main/java/org/hawkular/inventory/rest/RestRelationships.java
```diff
@@ -0,0 +1,150 @@
...
+
+/**
+ * @author jkremser
+ * @since 0.0.2
+ */
+@Path("/relationships")
+@Produces(APPLICATION_JSON)
+@Consumes(APPLICATION_JSON)
+@Api(value = "/relationships", description = "REST service for relationships")
+public class RestRelationships {
```
metlos, line 57: I'm not sure if this isn't more misleading than useful.

Since this is a top level endpoint, I'd assume that we'd be able to talk about relationships themselves, but instead we're only able to talk about relationships on tenants here.

jkremser, line 57: Relationships of all tenants. It should capture all rels except those between resource and it's type (same for metric). 
I agree, it is misleading. I wanted to provide a way to address the relationship by URL. It's both, Linked Data and HATEOASS, compliant. I guess, I could do that differently (with full path leading to the relationship, foo/inventory/relationships/1337 would be nice though)

btw. here is the JSON of a relation:

``` JSON
{
  "@context": {
    "inv": "http://hawkular.org/inventory/0.0.1",
    "baseUrl": "http://127.0.0.1:8080/hawkular/inventory/"
  },
  "@id": "baseUrl:?",
  "inv:shortId": "2",
  "@type": "inv:Relationship",
  "inv:source": {
    "@id": "baseUrl:tenants/acme",
    "@type": "inv:Tenant",
    "inv:shortId": "acme"
  },
  "inv:label": "contains",
  "inv:target": {
    "@id": "baseUrl:acme/resourceTypes/URL",
    "@type": "inv:ResourceType",
    "inv:shortId": "URL"
  },
  "inv:properties": {

  }
}
```

I was talking about the ? in the JSON - the addressability of 1 edge

metlos, line 57: I thought edges have their unique IDs?

Btw. I think it would be worth bringing this format to the dev list for discussion. While I think it is great that we are re-using a standard, I am not sure if we are supporting all the scenarios that we should in an ideal fashion. Or rather I am not even sure about the scenarios and queries that we need to support, so a discussion on the dev list I think would be beneficial.

metlos:
Bump. Where do we stand with this? (#77 is probably going to break this slightly).

jkremser:
closing this PR on behalf of #88

PR #57, opened 2015-04-27
[HWKINVENT-33] Paging for the REST API

metlos:


rest-servlet/src/main/java/org/hawkular/inventory/rest/RequestUtil.java
```diff
@@ -0,0 +1,66 @@
...
+public class RequestUtil {
+
+    private RequestUtil() {
+    }
+
+    public static Pager extractPaging(UriInfo uri) {
+        MultivaluedMap<String, String> params = uri.getQueryParameters();
+
+        String pageS = params.getFirst("page");
+        String perPageS = params.getFirst("per_page");
```
pilhuhn, line 41: While I saw that from the other guy. Is that `per_page` really common? Or should we consider `pageSize` or `page_size`?

rest-servlet/src/main/java/org/hawkular/inventory/rest/RestEnvironments.java
```diff
@@ -57,14 +60,16 @@
 
     @GET
     @Path("/{tenantId}/environments")
-    @ApiOperation("Returns all environments under given tenant.")
+    @ApiOperation("Returns all environments under given tenant. Accepts paging query parameters.")
```
pilhuhn, line 63: Is there a way to add them relatively generic to the api docs, as the signature does not explicitly expose them?

pilhuhn:
Can you please pull out the generic classes like `Page`, `Link`, `RequestUtil` etc. into a maven module that other (Hawkular-*) projects can also use? 
Those could live in the Kettle I think and be provided once by the Kettle, so that all projects can consider them as provided. But it may be safe to first ask @jmazzitelli for the best location of that common module.
And after that has happened, please talk about that module and its usage on hawkular-dev or better write a page for the developer guide on www.hawkular.org and link to it from the hawkular-dev list.

rest-servlet/src/main/java/org/hawkular/inventory/rest/RestEnvironments.java
metlos, line 63: Yeah, this is not ideal. I meant to find some "high level" place where we would spec out the params once and only reference them like this in the `@ApiOperation` docs. I obviously forgot to do that ;)

metlos:
:+1: I just wanted to give something out quickly now so that it can be discussed like this. The move can be done at any point after that.

rest-servlet/src/main/java/org/hawkular/inventory/rest/RestEnvironments.java
pilhuhn, line 63: On 28 Apr 2015, at 10:59, Lukas Krejci wrote:

> Yeah, this is not ideal. I meant to find some "high level" place where 
> we would spec out the params once and only reference them like this in 
> the `@ApiOperation` docs. I obviously forgot to do that ;)

What if we put them in a page on hawkular.org and add a link here like

```
 @ApiOperation("Returns all environments under given tenant. Accepts 
```

paging query parameters. See 
http://hawkular.org/dev/rest-paging_params")

rest-servlet/src/main/java/org/hawkular/inventory/rest/RequestUtil.java
jkremser, line 41: +1 for `per_page` GitHub uses it https://developer.github.com/v3/#pagination

rest-servlet/src/main/java/org/hawkular/inventory/rest/RestEnvironments.java
jkremser, line 63: What about `@QueryParam`s instead of extracting it from `UriInfo`?

pilhuhn, line 63: On 28 Apr 2015, at 11:51, Jirka Kremser wrote:

> > @@ -57,14 +60,16 @@
> > 
> >   @GET
> >   @Path("/{tenantId}/environments")
> > -    @ApiOperation("Returns all environments under given tenant.")
> > -    @ApiOperation("Returns all environments under given tenant. 
> >   Accepts paging query parameters.")
> 
> What about `@QueryParam`s instead of extracting it from `UriInfo`?

While we use that approach in RHQ, I see the benefit of not constantly 
repeating
ourselves via the solution from @metlos

metlos, line 63: yeah, when I thought I'd have to do the `@QueryParam` dance like 30 times, I was desperate to find another way ;)

metlos, line 63: Also, how do you support multiple values passed as via one parameter name (i.e. sort=a&sort=b) using `@QueryParam`?

pilhuhn, line 63: On 28 Apr 2015, at 14:52, Lukas Krejci wrote:

> > @@ -57,14 +60,16 @@
> > 
> >   @GET
> >   @Path("/{tenantId}/environments")
> > -    @ApiOperation("Returns all environments under given tenant.")
> > -    @ApiOperation("Returns all environments under given tenant. 
> >   Accepts paging query parameters.")
> 
> Also, how do you support multiple values passed as via one parameter 
> name (i.e. sort=a&sort=b) using `@QueryParam`?

This could be a matrix param sort=name=asc;street=desc;... and so on

metlos, line 63: Interesting. Never seen this before.

jkremser, line 63: > Also, how do you support multiple values passed as via one parameter name (i.e. sort=a&sort=b) using @QueryParam?

`@QueryParam List<Foo> foos`
Agreed, the code wouldn't be as DRY as it is, but it could be done also by some util class like this:

``` java

  getSomething(...., 
          @QueryParam("page") String page, 
          @QueryParam("per_page") String perPage, 
          @QueryParam("sort") List<String> sort, 
          @QueryParam("order") String order) {
      Pager pager = RequestUtil.extractPaging(page, perPage, sort, order);
      ...
  }
```

PR #59, opened 2015-04-30
Inspect by path

metlos:
Introduce a new way of traversing to a specific entity by specifying its class and the ids starting at tenant down the hierarchy to the entity.

This will prove useful in REST and other places that can compose the traversals more succinctly.


metlos:
This probably should reuse the `CanonicalPath` introduced in #77 (once that lands to master). I'm leaving this open just to remind us there's something to be done about this, but I don't consider this too much of a priority atm.

metlos:
Superseded by #89, which is now in master. Closing.

PR #63, opened 2015-05-18
[HWKINVENT-36]  Accounts Integration

metlos:
Before accounts could be reasonably integrated a couple of biggish refactorings had to be made so that it all made sense in the end.

Commit a22a6b12c2c165e706a9f6ff555cf1db4d765e8a:
- adds ability to filter the results and retain correct paging inside the implementations of inventory API (this is for now not used at all in the end, because it is only useful in a fine-grained security model we currently don't have).
- Add basic set of security operations and checks against them in the REST API.

Commit 09c74d119786d2489d07133c2213dc928c960058:
- introduces a concept of "autotenant" inventory - which creates a tenant on the fly even during the "get" calls. This is meant for the REST API supporting multitenancy so that we don't have to listen on a new user creation and pre-create a tenant for them. Instead, the tenant is created and initialized the first time it is asked for.
- Refactors how modified inventories are produced. Before the only modification we had was "observable inventory", now we also have "autotenant". The concept of "mixins" (well, hardcoded to be just the two aforementioned to keep strongly typed defs) is introduced so that one can create an inventory instance modified with any mixture of the modifications applied on top of it.
- The refactoring moved around some files so that we can have package-private access to classes and methods that should not be publicly consumable - namely observable inventory should only be obtainable using the mixin mechanism, not separately.

Commit bcbcd9bbad2471d5ceaf0cb4701fb0cb1475eb7a:
- restructures how bus is integrated into rest api war. This is now inline with our arch discussions - bus integration is achieved through an overlay over the "raw" rest api war.
- This commit also introduces the `TemporaryHacks` class which is used to auto-setup the newly created tenants with the test environment and types necessary for the pinger.
- Security operations are assigned to roles so that the actual perms checks work as intended.

Commit 2854747b22f4b37aaff317bb8cb4c050a51059b2:
- Final touches on the security operations.


metlos:
Also note that this renders the REST API integration tests non-functional. I will add further commits to this or another PRs to re-enable (and hopefully automate) them.

hawkular-integrated-inventory-rest/src/main/java/org/hawkular/integrated/inventory/TemporaryHacks.java
```diff
@@ -0,0 +1,69 @@
...
+import javax.enterprise.event.Observes;
+
+import static org.hawkular.inventory.api.Action.created;
+
+/**
+ * @author Lukas Krejci
+ * @since 0.0.2
+ */
+@ApplicationScoped
+public class TemporaryHacks {
```
jkremser, line 40: :sparkling_heart:

hawkular-inventory-cdi/src/main/java/org/hawkular/inventory/cdi/Observable_AutoTenant.java
```diff
@@ -0,0 +1,33 @@
...
+import java.lang.annotation.Target;
+
+/**
+ * @author Lukas Krejci
+ * @since 0.0.2
+ */
+@Retention(RetentionPolicy.RUNTIME)
+@Target({ElementType.FIELD, ElementType.PARAMETER, ElementType.METHOD})
+@Qualifier
+public @interface Observable_AutoTenant {
```
jkremser: Why the underscore? `@Observable_AutoTenant` vs `ObservableAutoTenant`

impl-tinkerpop-parent/impl/src/main/java/org/hawkular/inventory/impl/tinkerpop/TenantBrowser.java
```diff
@@ -72,7 +74,7 @@ private TenantBrowser(InventoryContext context, FilterApplicator.Tree path) {
             }
 
             @Override
-            public Tenant entity() {
+            public Tenant entity() throws EntityNotFoundException, RelationNotFoundException {
```
jkremser, line 77: So `{Entity|Relation}NotFoundException` are now checked, or is it just about being more explicit?

hawkular-inventory-cdi/src/main/java/org/hawkular/inventory/cdi/Observable_AutoTenant.java
metlos: I honestly don't know ;) This probably should be changed.

impl-tinkerpop-parent/impl/src/main/java/org/hawkular/inventory/impl/tinkerpop/TenantBrowser.java
metlos, line 77: Just more explicit...

metlos:
I've removed the "Don't Merge Yet" label. The only thing that will be left broken are the integration tests but those will need more discussion across other hawkular components. So let's just not wait for that to pan out and get this in if no other problems with the PR itself are found.

PR #65, opened 2015-05-20
Tenant in headers

metlos:
Following the suite of metrics, the tenant is now read from the headers, or rather using accounts in whatever way it chooses.

### DIFF
--- a/rest-servlet/src/main/java/org/hawkular/inventory/rest/RestEnvironments.java
+++ b/rest-servlet/src/main/java/org/hawkular/inventory/rest/RestEnvironments.java
@@ -56,44 +56,51 @@ import static org.hawkular.inventory.rest.ResponseUtil.pagedResponse;
 public class RestEnvironments extends RestBase {
 
     @GET
-    @Path("/{tenantId}/environments")
+    @Path("/environments")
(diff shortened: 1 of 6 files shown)


jkremser:
Once this is merged it'd break the UI. I can prepare the UI for this change or ping someone, before merging.

metlos:
UI depends on inventory 0.0.9 while this will go into 0.1.0. We can merge now and prepare the UI later. It would be good to create a UI JIRA for that though...

jkremser:
It looks good, providing the following code returns really the current persona and not some stale data:

``` java
RestBase:
    protected String getTenantId() {
        return personas.getCurrent().getId();
    }
```

EDIT:
it takes the current persona from the (hopefully)current http request

``` java
String personaId = this.httpRequest.getHeader("Hawkular-Persona")
```

jkremser:
> UI depends on inventory 0.0.9 while this will go into 0.1.0. We can merge now and prepare the UI later. It would be good to create a UI JIRA for that though...

right

metlos:
As for obtaining the current persona - there are more ways of doing that in accounts, so we can choose another if this somehow breaks. Btw. the `getTenantId()` method is already in 0.0.9 with the same impl.

PR #68, opened 2015-05-22
HAWKULAR-242 - Added @Singleton to Security

jpkrohling:


jpkrohling:
This seems to fix the issue, but please, make sure that Security indeed can be made Singleton. I checked for references, and all I could find was one place that injected Security, so, it should be safe. But would be nice to confirm that it's also OK for possible other places consuming this class.

jpkrohling:
@theute , @pilhuhn , @ammendonca , could one of you please test this fix? You were able to reproduce it quite consistently, while I was able to reproduce only after removing the data/tmp directories. In my tests, it seemed to work.

How to test it:
- build inventory with this patch (mvn clean install -DskipTests)
- copy hawkular-inventory/hawkular-integrated-inventory-rest/target/hawkular-integrated-inventory-rest-0.0.6-SNAPSHOT.war as hawkular/dist/target/hawkular-1.0.0-SNAPSHOT/wildfly-8.2.0.Final/modules/org/hawkular/nest/main/deployments/hawkular-integrated-inventory-rest.war
- boot hawkular

metlos:
Not sure about the interactions with the permission checker and others that are injected into this, but the inventory injection should be safe as the inventory is a singleton itself.

ammendonca:
@jpkrohling just tested this and indeed it seems to fix the issue, URLs can be added and list is shown. No metrics are gathered yet, but this is already helpful. Thanks.

PR #69, opened 2015-05-23
Pinger fixes

metlos:
Miscellaneous fixes related to making pinger work.

With this, adding and monitoring URLs using pinger works. Delete is still problematic (read broken) and will be fixed in a separate PR.


metlos:
Emergency merge to get this into repos so that others can start working with pinger on Monday.

PR #71, opened 2015-05-26
add more units - match what wildfly and rhq support

jmazzitelli:


api/src/main/java/org/hawkular/inventory/api/model/MetricUnit.java
```diff
@@ -26,13 +26,54 @@
...
     SECONDS("s"),
-    MINUTE("min"),
-    BYTE("b"),
-    KILO_BYTE("kb");
+    MINUTES("m"),
+    HOURS("h"),
+    DAYS("d"),
+
+    // Rate
+    PER_JIFFY("/j"),
```
metlos, line 64: Are these really useful? If we are going to support derived units, we should maybe support them a little bit more elaborately than this.

pilhuhn, line 64: Jiffy is a strange thingy anyway :)

There is JSR 363 that may be helpful here.

pilhuhn:
I wonder a) how much to align things here with JSR 363 and b) how much of this can/has to be shared with -metrics, as at one point in time they also want to tag metrics with units and the logic to up- and down-scale them.

jmazzitelli:
Note that these come straight from WildFly MeasurementUnit. If we want to support a subset of them, that's OK. I just figured we want to at least match the names of what is there (even if its a subset).

It will make mapping between them easy:

MeasurementUnit.valueOf(inventoryMetricUnit.name)

and vice versa.

metlos:
I'm a little bit concerned about using JSR 363 right now, because the impl  is not GA and the standard has not been approved yet.

The implementation is here: https://github.com/unitsofmeasurement/uom-se
The API is here: https://github.com/unitsofmeasurement/unit-api
and Jackson (de)serializer is here: https://github.com/unitsofmeasurement/jackson-module-unitsofmeasure

All of them seem to be at various levels of completion.

I think I'm inclined to just include Wfly's units as @jmazzitelli proposes, mainly out of the fear of unknown ;)

That being said, if we decide to go down that route, could you please make this pass the CI, Mazz?

jmazzitelli:
> That being said, if we decide to go down that route, could you please make this pass the CI, Mazz?

I just saw this now. This is no longer easily merged because of renamed/moved directories. I will create another PR with this same diff but based on master and close this one.

jmazzitelli:
see PR#145 as the replacement to this one

PR #75, opened 2015-05-28
HWKINVENT-53 using the uuid instead of URL itself if the URL is longe…

jkremser:
…r than 250 characters.

Also adding some unit tests.

### DIFF
--- a/rest-servlet/src/main/java/org/hawkular/inventory/rest/Security.java
+++ b/rest-servlet/src/main/java/org/hawkular/inventory/rest/Security.java
@@ -45,6 +45,7 @@ import javax.transaction.UserTransaction;
 import java.util.EnumMap;
 import java.util.HashMap;
 import java.util.Map;
+import java.util.UUID;
 
 /**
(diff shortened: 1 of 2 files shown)


rest-servlet/src/main/java/org/hawkular/inventory/rest/Security.java
```diff
@@ -152,7 +153,7 @@ public static String getStableId(Class<? extends AbstractElement<?, ?>> type, St
                 return join(ids[0], ids[1], ids[2], "metrics", ids[3]);
             }
         } else if (Relationship.class.isAssignableFrom(type)) {
-            return join("relationships", ids[0]);
+            return "relationships/" + ids[0];
```
jkremser, line 156: this is here because of this https://github.com/Jiri-Kremser/hawkular-inventory/blob/a0d8919e2d9715654d5041ced2daac96a6ce99a8/rest-servlet/src/main/java/org/hawkular/inventory/rest/Security.java#L165

PR #76, opened 2015-05-28
HAWKULAR-255 Adding the DuplicateFilterPipe to the end of the query pipe line.

jkremser:
This is not optimal, because now each query needs to run the dedup step. Internally, it uses `LinkedHashSet` so it is in `O(n)`


impl-tinkerpop-parent/impl/src/main/java/org/hawkular/inventory/impl/tinkerpop/FilterApplicator.java
```diff
@@ -91,7 +91,8 @@ private FilterApplicator(Type type, T f) {
                 branches.add(branch);
             }
 
-            q.copySplit(branches.toArray(new HawkularPipeline[branches.size()])).exhaustMerge();
+            // HAWKULAR-255: the dedup pipe in the end could go away once we find out why there are duplicates
```
metlos, line 94: Please add a TODO here so that IDEs flag this up. We need to take another look at this.

PR #77, opened 2015-05-29
[HWKINVENT-57] Move business logic to base impl inside API

metlos:
This is a big, big refactoring.
## What's wrong

There is a support for listening on delete events in the inventory. The problem with it is that only the entity being explicitly deleted is reported in the event. Any entities that are deleted along with it due to them being contained in the deleted entity (i.e. there being a `contains` relationship from the deleted entity to the contained entities, recursively) are NOT reported as such.

The only chance for the user of the inventory to figure out what will be deleted is to manually traverse the transitive closure over the contains edges from the entity to be deleted PRIOR to it being deleted. This is sub-optimal to say the least.
## How To Fix It

I've meant to do it for some time, but there was never a compelling reason to do it until now. The inventory as it exists at the moment implements large parts of the "business logic" (i.e. the semantic rules for different relationships, query composition, etc.) inside a concrete implementation that deals with a particular backend. But should we ever support another backend (like tinkerpop3 that we will have to support in order to use the future Titan 1.0), that logic will have to be duplicated in that implementation, too.

So the fix is to pull as much of the implementation of the API out of the tinkerpop impl into the API itself, currently into a package called "lazy" to suggest that the backend is queried lazily only when some actual results are needed and not during the individual "hops" through the inventory.

This will leave a rather small SPI interface that the real "backends" will have to implement.

Along with the change I also pulled out the actual test suite from the tinkerpop impl up into the API tests. A concrete backend implementation will then depend on the test-jar of the API and will inherit this abstract testsuite providing the implementation to test with otherwise standard test suite.

Once we have common business logic and behavior up in the API, we can easily add support for sending events during the delete without a fear that some other implementation will "forget" to do it because it will be the API itself that will be doing it.
## Progress
- [x] Pull out the impl
- [x] Make the testsuite pass
- [x] Rename "lazy" to something more meaningful like "ri" (reference impl) and strongly suggest using it instead of reimplementing the whole API.
- [x] Javadocs, javadocs and more javadocs.
- [x] Tests of the backend SPI itself.
- [x] Remove the "old" implementation from the Tinkerpop module.
- [x] Merge `ObservableInventory` into the main API
- [x] Fire additional events during the delete
- ~~[ ] Port the fix for HWKINVENT-255.~~ This needs to be revisited anyway once we start running on Titan in Hawkular.


api/src/main/java/org/hawkular/inventory/lazy/Mutator.java
```diff
@@ -0,0 +1,346 @@
...
+import static org.hawkular.inventory.api.filters.Related.by;
+import static org.hawkular.inventory.api.filters.With.id;
+import static org.hawkular.inventory.api.filters.With.type;
+
+/**
+ * @author Lukas Krejci
+ * @since 0.0.6
+ */
+abstract class Mutator<BE, E extends Entity<Blueprint, Update>, Blueprint extends Entity.Blueprint,
+        Update extends AbstractElement.Update> extends Traversal<BE, E> {
```
jkremser: `abstract class Mutator<BE, E extends Entity<Blueprint, Update>, Blueprint extends Entity.Blueprint, Update extends AbstractElement.Update> extends Traversal<BE, E>` hmmm :]

metlos: Yeah, I know... javadoc would help here as would saner generics syntax ;)

metlos:
I need to rebase all the commits to rename the [HWKINVENT-45] under which I originally started fixing this to [HWKINVENT-57] which the fix ended up being implemented under.

impl-tinkerpop-parent/impl/src/main/java/org/hawkular/inventory/impl/tinkerpop/TinkerpopBackend.java
```diff
@@ -0,0 +1,728 @@
...
+            throw new ElementNotFoundException();
+        } else {
+            return it.next();
+        }
+    }
+
+    @Override
+    public Page<Element> query(Query query, Pager pager) {
+        HawkularPipeline<?, ? extends Element> q;
+        if (query.getFragments()[0].getFilter() instanceof RelationFilter) {
```
jkremser, line 95: This looks more readable than the `JumpInOutFilter` filter. What is the `SwitchElementType` good for, it looks like the same thing as the old `JumpInOutFilter`, but here I can see it's not used anymore.

api/src/main/java/org/hawkular/inventory/base/Mutator.java
```diff
@@ -0,0 +1,431 @@
...
+            context.backend.commit();
+        } catch (Exception e) {
+            context.backend.rollback();
+            throw e;
+        }
+
+        //report the relationship deletions first - it would be strange to report deletion of a relationship after
+        //reporting that an entity on one end of the relationship has been deleted
+        for (Relationship r : deletedRelationships) {
+            context.notify(r, deleted());
```
jkremser, line 283: Isn't entity-based notification enough? Perhaps adding/removing an edge can be considered also as the entity update, I don't know.

api/src/main/java/org/hawkular/inventory/base/Mutator.java
```diff
@@ -0,0 +1,431 @@
...
+            if (parentPath.path.isDefined()) {
+                BE parent = parentPath.entity;
+                containsRel = context.backend.relate(parent, entityObject, contains.name(), Collections.emptyMap());
+            }
+
+            newEntity = wireUpNewEntity(entityObject, blueprint, parentPath.path, parentPath.entity);
+
+            context.backend.commit();
+        } catch (Throwable t) {
+            context.backend.rollback();
```
jkremser, line 109: where does the transaction start? I recall some chat about it on IRC, but if it is the first call to the backend after previous commit/rollback, it's somewhat strange, because the read only calls don't necessarily need to open the transaction.

Is the graph smart enough to recognize `not-modifying op` and not open the transaction for them?

impl-tinkerpop-parent/impl/src/main/java/org/hawkular/inventory/impl/tinkerpop/TinkerpopBackend.java
metlos, line 95: It is the same thing as `JumpInOutFilter`. It is used in the base impl, afair. The backend just translates the Query to its native form - here a gremlin pipeline and executes it. There's still the `FilterApplicator` which mostly remained the same where you can see how it applies the `SwitchElementType` (which as you correctly found out is just a renamed `JumpInOutFilter`).

api/src/main/java/org/hawkular/inventory/base/Mutator.java
metlos, line 283: Well, I think people would be interested if two entities were no longer related. If you allow for relationship notifications, you need to be consistent and therefore emit them also during entity deletions and creations.

api/src/main/java/org/hawkular/inventory/base/Mutator.java
metlos, line 109: Transactions in blueprints are thread-local and start with the first access to the database (read or write).
Which reminds me - we need to do `commit()`s even in the querying calls because otherwise we risk seeing stale data.

api/src/main/java/org/hawkular/inventory/base/Mutator.java
pilhuhn, line 283: > Well, I think people would be interested if two entities were no 
> longer related. If you allow for relationship notifications, you need 
> to be consistent and therefore emit them also during entity deletions 
> and creations.

I agree. In the future we may have a UI that graphically shows 
components
of an app and potentially relations to other apps. So if you e.g. show
app as lb+3*as+db and have a link from db to other apps, you want
that graph get updated if one of the links from db to other apps goes
away.

PR #78, opened 2015-06-01
HAWKULAR-255 Adding synchronization for modifying operations of vertex.

jkremser:
After this change I wasn't able to replicate the issue, hopefully it's fixed now.


metlos:
Good job tracking this down!

PR #79, opened 2015-06-03
Removing dependencies to Scala by using the plain Java swagger libraries

jkremser:
I'd like to test it on inventory + travis + gh pages before moving the deps to the parent pom.


metlos:
Given it passes CI, can this be merged or do you want to perform some Travis-y magic on it?

jkremser:
It can be merged. I am locally working on `swagger-maven-plugin` upgrade, they completely broke everything so the mustache template needs to be changed as well. I'll probably create PR with this change directly to the build tools repo once it's done

PR #81, opened 2015-06-10
[HWKINVENT-58] Port the fix for HAWKULAR-255 to the new inventory implementation.

metlos:
The nice thing about this approach is that it is both specific to
tinkergraph and doesn't lay additional burden on the Titan impl and it
also is future-proof because it will be possible to use this also with
Tinkerpop3 and its explicit transaction handling.

### DIFF
--- a/impl-tinkerpop-parent/tinkergraph-provider/src/main/java/org/hawkular/inventory/impl/tinkerpop/provider/TinkerGraphProvider.java
+++ b/impl-tinkerpop-parent/tinkergraph-provider/src/main/java/org/hawkular/inventory/impl/tinkerpop/provider/TinkerGraphProvider.java
@@ -67,14 +67,20 @@ public final class TinkerGraphProvider implements GraphProvider<TinkerGraphProvi
 
     @Override
     public void commit(WrappedTinkerGraph graph, InventoryBackend.Transaction t) {
-        graph.commit();
-        unlock(t);
+        try {
+            graph.commit();
+        } finally {
+            unlock(t);
+        }
     }
 
     @Override
     public void rollback(WrappedTinkerGraph graph, InventoryBackend.Transaction t) {
-        graph.rollback();
-        unlock(t);
+        try {
+            graph.rollback();
+        } finally {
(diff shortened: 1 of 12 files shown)


impl-tinkerpop-parent/tinkergraph-provider/src/main/java/org/hawkular/inventory/impl/tinkerpop/provider/TinkerGraphProvider.java
```diff
@@ -39,6 +45,47 @@ public void ensureIndices(WrappedTinkerGraph graph, IndexSpec... indexSpecs) {
...
+            lock.readLock().lock();
+        }
+
+        return new SimulatedSerializedTransaction(mutating, lock);
+    }
+
+    @Override
+    public void commit(WrappedTinkerGraph graph, InventoryBackend.Transaction t) {
+        graph.commit();
+        unlock(t);
```
jkremser: perhaps

``` java
try {
  graph.commit();
} finally {
  unlock(t);
}
```

.. would be better

What if I start a transaction and forget to call the commit/rollback (or I forgot to place it to the finally clause and exception is thrown). We'll get the deadlock, right? Doing some timeout logic is overkill for the TinkerGraph, perhaps we'll get it for free with the Titan.

metlos: If you look at the implementation of `WrappedTinkergraph#commit()`, you'll notice it is a no-op. But your solution is technically more correct. I'll update the code.

The assumption with the locks is that the user (the base impl) does it right. I would not go further with Tinkergraph impl - timeouts would complicate the matters greatly because they could corrupt the database in the same way it could be corrupted prior to HAWKULAR-255.

Titan and any other implementation has the opportunity to do this differently and in fact the default implementations of the methods on the interface actually use no locks at all because they assume the underlying graph is truly transactional (which TinkerGraph is not).

PR #82, opened 2015-06-10
HWKINVENT-48 Create feed API should return error if feed ID already exists

jkremser:
Throwing an exception when registering with the same feed id more than once.

Good practice from the relation db world is to set the id to null and let the backend generate the id. However, it's not possible to create the `Feed` instance with `null` id, because there is a check in the `AbstractElement` class, so I was little bit cheating and set some predefined string there when calling `new Feed(null)` to determine if the id generation is wanted.

### DIFF
--- a/api/src/main/java/org/hawkular/inventory/api/FeedAlreadyRegisteredException.java
+++ b/api/src/main/java/org/hawkular/inventory/api/FeedAlreadyRegisteredException.java
@@ -1,0 +1,38 @@
+/*
+ * Copyright 2015 Red Hat, Inc. and/or its affiliates
+ * and other contributors as indicated by the @author tags.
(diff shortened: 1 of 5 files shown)


api/src/main/java/org/hawkular/inventory/api/FeedAlreadyRegisteredException.java
```diff
@@ -0,0 +1,38 @@
...
+ * limitations under the License.
+ */
+
+package org.hawkular.inventory.api;
+
+import org.hawkular.inventory.api.model.Feed;
+
+/**
+ * @author jkremser
+ * @since 1.0
```
metlos: 0.1.0

PR #84, opened 2015-06-10
[HWKINVENT-59] Implement filter for entity property values.

metlos:
Also consolidated the naming and capabilities of the similar filter on
relationships.

### DIFF
--- a/api/src/main/java/org/hawkular/inventory/api/filters/RelationWith.java
+++ b/api/src/main/java/org/hawkular/inventory/api/filters/RelationWith.java
@@ -37,30 +37,34 @@ public final class RelationWith {
         return new Ids(ids);
     }
 
-    public static Properties property(String property, String value) {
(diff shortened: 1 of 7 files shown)


api/src/main/java/org/hawkular/inventory/api/filters/RelationWith.java
```diff
@@ -37,30 +37,34 @@ public static Ids ids(String... ids) {
         return new Ids(ids);
     }
 
-    public static Properties property(String property, String value) {
```
jkremser, line 40: The PR looks good to me. However, if I merge it the rest-servlet module won't compile. Could you please rebase it against the latest master. I think the only problem will be on this line http://git.io/vLw7E : 

`property(String, String)` -> `propertyValue(String, String)` or even something more clever, like if the value is an empty string, then use the `property(String)` which, I assume, checks if the property is there)

PR #86, opened 2015-06-12
HWKINVENT-41 Pull out the CORS filter

ppalaga:


ppalaga:
@metlos I hope this is complete. I did some rudimentary checks if CorsFilter is referenced somewhere else but have not found anything.

PR #87, opened 2015-06-16
HWKINVENT-72 and HWKINVENT-49

ppalaga:
Integration tests adapted in https://github.com/hawkular/hawkular/pull/214


metlos:
Great job! Now don't look like the left hand doesn't know what the right hand does anymore ;)

PR #88, opened 2015-06-16
Hwkinvent 18 - exposing relationships

jkremser:


api/src/main/java/org/hawkular/inventory/base/spi/CanonicalPath.java
```diff
@@ -113,6 +114,32 @@ public CanonicalPath visitUnknown(Object entity, Void parameter) {
         }, null);
     }
 
+    public Entity toEntity() {
```
metlos, line 117: This creates basically invalid entities - like resource without a resource type or metric type without a unit or indeed a resource type with hardcoded version to "1.0".

rest-servlet/src/main/java/org/hawkular/inventory/rest/RestResources.java
```diff
@@ -51,21 +45,28 @@
...
 
 /**
  * @author Lukas Krejci
  * @since 1.0
  */
 @Path("/")
-@Produces(value = APPLICATION_JSON)
-@Consumes(value = APPLICATION_JSON)
+@Produces(APPLICATION_JSON)
+@Consumes(WILDCARD)
```
metlos: Why wildcard here and not anywhere else?

api/src/main/java/org/hawkular/inventory/base/spi/CanonicalPath.java
jkremser, line 117: I know, but it's ok for it's purpose, i.e. to create the right entity type with the right id, which goes to the `new Relationship()` ctor. Any other idea how to de-serialize a relationship from json? Finding the right source/target in the db is imho over-kill.

rest-servlet/src/main/java/org/hawkular/inventory/rest/RestTenants.java
```diff
@@ -100,4 +134,20 @@ public Response delete() {
...
+    @DELETE
+    @Path("/")
+    @ApiOperation("Deletes the tenant and all its data. Be careful!")
+    @ApiResponses({
+                          @ApiResponse(code = 204, message = "OK"),
+                          @ApiResponse(code = 401, message = "Unauthorized access"),
+                          @ApiResponse(code = 404, message = "Tenant doesn't exist", response = ApiError.class),
+                          @ApiResponse(code = 500, message = "Server error", response = ApiError.class)
+                  })
+    public Response delete(@PathParam("tenantId") String tenantId) {
```
metlos: `tenantId` is not mentioned in the `@Path` annotation on the method. Is it correct to mention it here? Also for consistency with `GET /tenants/{tenantId}`, should this be `@Path("/tenants/{tenantId}")`?

rest-servlet/src/main/java/org/hawkular/inventory/rest/RestResources.java
jkremser: good catch, I was trying to fix an issue\* with keycloak and forgot on this one.

(*) the api doesn't work if the accept header is not correctly set, that's why our API is not browsable in a browser (browser doesn't set the application/json accept header).

api/src/main/java/org/hawkular/inventory/base/spi/CanonicalPath.java
metlos, line 117: Yeah, once my changes for the canonical path are in, I'm actually thinking of using that instead of the full entity references in a lot of places - like relationship targets for example...

We can leave this in for the moment, but I think I will redo this once the resource hierarchy and canonical path changes are in. This will become much simpler.

rest-servlet/src/main/java/org/hawkular/inventory/rest/RestTenants.java
jkremser: right, I missed that, the parameter is not used at all

> Also for consistency with GET /tenants/{tenantId}, should this be @Path("/tenants/{tenantId}")?

I wanted the URI in the relationship json to be the real URL, but perhaps it's not a good idea, because it's not only the tenants.. All the endpoints would need to have some kind of shadow implementation to be responding also for the old form of the api like:
- `28026b36-8fe4-4332-84c8-524e173a68bf/metricTypes/status.code.type`
- `28026b36-8fe4-4332-84c8-524e173a68bf/resourceTypes/URL`
- etc.

Using the URI without the tenantId is not good idea either, because of the uniqueness, perhaps the not meaningful URIs are just fine. Not very HATEOAS, but better than adding more complexity to the rest classes.

metlos: Good. We could use the serialized canonical path form from my upcoming changes then..

Both for stable ids for authz and for json-ld stuff.

I think having a real URL in json-ld is still somewhat desired so we may introduce a new endpoint `/element?id=...` where the `...` would be the serialized form of the canonical path.

The serialized form right now looks like:
`t|tenantId/e|envId/f|feedId/r|parentResource/r|childResource`. We could change that so that it is more easily passed as a query parameter (not sure if the slashes would cause problem in a query param value or not).

metlos:
K, please remove the `WILDCARD` and maybe even comment out the changed tenant endpoints and I think this is ready to be merged. Great job!

PR #89, opened 2015-06-17
[HWKINVENT-10] Resource hierarchy

metlos:
This is rather big (again) but I think provides a good foundation for future changes.
- "owns" relationship renamed to "incorporates" - the name was confusing
  for what the relationship really meant
- Semantical rules for relationship creation/deletion are checked for
  throughout the codebase - it is no longer possible to "manually" delete
  the contains relationship and wreck havoc in the inventory structure.
- All elements now store a canonical path in them. This makes the conversion
  of entities from internal representations much cheaper and also makes it
  possible to figure out the canonical path to a nested resource efficiently.
  Without it a full traversal from the resource up to tenant would be needed
  to obtain that information.
- "isParentOf" relationship added for obvious reasons given the purpose of
  this patch ;)
- [x] Java API updates
- [x] REST API updates


metlos:
This is now ready for review. My apologies for the size of this change. I yet again failed to slice it up into independent(ish) chunks, because the original feature - resource hierarchy wouldn't make much sense without the paths - which is dealt with in majority of the commits.

I do promise to try to make a better job of separating the features and making the PRs smaller.

api/src/main/java/org/hawkular/inventory/api/model/Path.java
```diff
@@ -0,0 +1,980 @@
...
+    /**
+     * If this path was created by going {@link #up() up} from another path, then this method can be used to go back
+     * down n steps to the previous paths representing some (grand-)child resource of the resource represented by
+     * this path object.
+     *
+     * @param distance the distance from this path to the child path
+     * @return a path to a child of the resource represented by this path (may be {@link #isDefined() undefined}.
+     */
+    public Path down(int distance) {
+        return newInstance(startIdx, endIdx + distance, path);
```
jkremser, line 274: so calling `down()` without previously calling `up()` doesn't make sense, right?

metlos, line 274: Yes, the returned path will be undefined. Doing `path.down().up()` should return a path equivalent to the original.

api/src/main/java/org/hawkular/inventory/api/model/RelativePath.java
```diff
@@ -0,0 +1,542 @@
...
+        SHORT_TYPE_NAMES.put(Up.class, "..");
+
+        List<Class<?>> justUp = Collections.singletonList(Up.class);
+
+        VALID_PROGRESSIONS.put(Tenant.class, Arrays.asList(Environment.class, MetricType.class, ResourceType.class,
+                Up.class));
+        VALID_PROGRESSIONS.put(Environment.class, Arrays.asList(Metric.class, Resource.class, Feed.class, Up.class));
+        VALID_PROGRESSIONS.put(Feed.class, Arrays.asList(Metric.class, Resource.class, Up.class));
+        VALID_PROGRESSIONS.put(Resource.class, Arrays.asList(Resource.class, Up.class));
+        VALID_PROGRESSIONS.put(null, Arrays.asList(Tenant.class, Relationship.class, Up.class));
```
jkremser, line 73: what this line exactly means? That it's possible to do these kind of progressions no matter what the current type is? What does `up()` on tenant do? identity?

api/src/main/java/org/hawkular/inventory/api/model/Path.java
jkremser, line 274: then perhpaps `back()` is better name, but it has it's drawbacks too..

api/src/main/java/org/hawkular/inventory/api/model/ResourceType.java
```diff
@@ -39,19 +40,19 @@
     private ResourceType() {
     }
 
-    public ResourceType(String tenantId, String id) {
-        super(tenantId, id);
+    public ResourceType(CanonicalPath path) {
+        super(path);
```
jkremser, line 44: takze tady ty zmeny tech signatur konstruktoru a na tech dalsich entitach to taky vynucuje pak tu zmenu v `hawkular/hawkular`

jkremser, line 44: kua, english

api/src/main/java/org/hawkular/inventory/base/RelationshipRules.java
```diff
@@ -0,0 +1,223 @@
...
+/**
+ * Some well-known relationships have certain semantic rules that need to be checked for when creating/deleting them.
+ *
+ * <p>This class concentrates those checks so that they can be called easily from the various places in the codebase
+ * that work with relationships.
+ *
+ * @author Lukas Krejci
+ * @since 0.2.0
+ */
+public final class RelationshipRules {
```
jkremser, line 44: nice!

impl-tinkerpop-parent/impl/src/main/java/org/hawkular/inventory/impl/tinkerpop/HawkularPipeline.java
```diff
@@ -71,6 +73,11 @@ public HawkularPipeline(Object starts, boolean doQueryOptimization) {
         super(starts, doQueryOptimization);
     }
 
+    public String nextRandomLabel() {
+        //exceptionally random, right? ;)
+        return ")(*)#(*&$(" + asLabelCount++;
```
jkremser, line 78: i'd add more "!!!" to make it more random ;)

rest-servlet/src/main/java/org/hawkular/inventory/rest/json/PathSerializer.java
```diff
@@ -0,0 +1,56 @@
...
+ * and shared and therefore cannot be used for a context-specific, and therefore stateful, work.
+ *
+ * @author Lukas Krejci
+ * @since 0.2.0
+ */
+public class PathSerializer extends JsonSerializer<CanonicalPath> {
+    @Override
+    public void serialize(CanonicalPath value, JsonGenerator jgen, SerializerProvider provider) throws IOException {
+        //strip the tenant from the canonical path so that it appears to users it doesn't even exist. They don't
+        //need to provide it in their paths and the API doesn't show it to them either.
```
jkremser, line 43: What is user wants to create a relationship between two entities from different tenants, how this is possible without identifying the tenant in the c. path? I guess, it's currently disallowed with those rel. checkers, but perhaps we will need it in the future (?) I don't know, tenant's data should be isolated, so I guess it's ok.

api/src/main/java/org/hawkular/inventory/api/model/Path.java
metlos, line 274: if `down()` was `back()` what would `up()` become?

The down and up names come from moving up and down a path.. I guess you could also move back and forth, but `path.forth()` doesn't make that much sense to me.

api/src/main/java/org/hawkular/inventory/api/model/RelativePath.java
metlos, line 73: Moving up from a tenant leaves you in the unknown. `null` on line 73 expresses that unknown, so if your path is in the unknown "realm", you can either go down to tenants or relationships or remain in the unknown using "up".

api/src/main/java/org/hawkular/inventory/api/model/ResourceType.java
metlos, line 44: jo :smile:

rest-servlet/src/main/java/org/hawkular/inventory/rest/json/PathSerializer.java
metlos, line 43: While it has been allowed to create cross-tenant rels in the past (mostly due to lack of validation in the first iterations of the impl), I think we should disallow that going forward. Tenants are meant to be a "unit of data separation" so enabling relationships between entities inside different tenants seems wrong to me from that POV.

PR #92, opened 2015-06-19
Upgrade to hawkular-parent 16 - upgarde to WildFly 9

ppalaga:


rest-servlet/src/main/java/org/hawkular/inventory/rest/RestFilters.java
```diff
@@ -0,0 +1,61 @@
+/*
```
metlos: Is this file supposed to be part of this PR?

ppalaga:
@metlos sorry based on an improper branch, sorry.

ppalaga:
e8006a2 should pass CI as I removed the outdated `jboss-javaee-7.0-with-resteasy` import from rest-api

ppalaga:
@metlos could you please merge?

PR #98, opened 2015-06-25
HWKINVENT-68 Removing Integration tests from Inventory after they have

ppalaga:
moved to Hawkular


ppalaga:
It was agreed with @metlos that we will remove the rest-tests module altogether from here once we move it to Hawkular main project, which happened recently.

PR #99, opened 2015-06-26
HWKINVENT-60 Add basic filtering in Inventory REST

ppalaga:


ppalaga:
The scope of HWKINVENT-60 was narrowed a little bit to cover only stuff that works with the present inventory backend and there is a new PR that addresses the new scope: https://github.com/hawkular/hawkular-inventory/pull/101 Thus closing this one.

PR #101, opened 2015-06-29
HWKINVENT-60 Add basic filtering in Inventory REST

ppalaga:


ppalaga:

rest-servlet/src/test/java/org/hawkular/inventory/rest/filters/ParamPropertiesTest.java
```diff
@@ -0,0 +1,97 @@
...
+    private void assertParse(String serialized, Object... keys) {
+        LinkedHashMap<String, List<String>> expected = new LinkedHashMap<>();
+        for (int i = 0; i < keys.length; ) {
+            String key = (String) keys[i++];
+            Object v = keys[i++];
+            if (v == null) {
+                expected.put(key, null);
+            } else {
+                String[] valArray = v instanceof String ? new String[] {(String) v} : (String[]) v;
+                List<String> valList = new ArrayList<>();
```
metlos, line 87: Nitpick: Wouldn't Arrays.asList() be simpler here?

metlos:
We need to start producing some docs around changes like this. Would you write some documentation for the types of filters accepted and the value format? Possibly both in javadoc and hawkular.org... May be good to track as a separate JIRA.

Otherwise good job. I like the extendability of this approach.

rest-servlet/src/test/java/org/hawkular/inventory/rest/filters/ParamPropertiesTest.java
ppalaga, line 87: Actually +1 here in this test. Otherwise, I tend to avoid `Arrays.asList()` because it returns a writable `java.util.Arrays.ArrayList` instance backed by the array. Because the array can leak, one cannot reliably make the ArrayList immutable. But you are right that it is not a problem here.

ppalaga:
Created a follow-up https://issues.jboss.org/browse/HWKINVENT-79 for the parts uncovered in the present PR.

PR #102, opened 2015-07-01
HWKINVENT-80 Add a backend test for filtering by properties of related

ppalaga:
entities


ppalaga:

PR #103, opened 2015-07-01
[HWKINVENT-83] Provide a more robust means for configuring inventory.

metlos:
This should enable us to share some of the configuration env variables
with Metrics.


ppalaga:
lgtm

PR #104, opened 2015-07-07
HWKINVENT-61 Adding new module with a load scenario defined in Gatling

jkremser:
todo:
- [x] ~~In the server.log I get nondeterministically these errors: "Resource to be checked is invalid (null)" Because the resource is not properly looked up here http://git.io/vqEHA (`HWKINVENT-85`)~~ PEBKAC :)
- [x] test adding resource types, metric types and associating them together
- [x] test adding resources, metrics and associating them together

Example report:
https://jiri-kremser.github.io/hawkular-inventory/agentsimulation-1436398368281

### DIFF
--- a/load-tests/README.adoc
+++ b/load-tests/README.adoc
@@ -1,0 +1,24 @@
+= Hawkular Inventory Load Tests
+
+A simple load testing tool for Hawkular Inventory
+
+== Why a Maven project?
+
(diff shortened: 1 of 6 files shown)


metlos:
I'll merge this in so that we can work with this further in master. Great job, Jirka, the reports are looking great. Hope they will enable us to improve on the performance.

PR #105, opened 2015-07-08
HWKINVENT-84 | default value of  MetricUnit is set to NONE instead of null

pavolloffay:


api/src/main/java/org/hawkular/inventory/api/model/MetricType.java
```diff
@@ -157,7 +155,7 @@ private Update() {
 
         public Update(Map<String, Object> properties, MetricUnit unit) {
             super(properties);
-            this.unit = unit;
+            this.unit = unit == null ? MetricUnit.NONE : unit;
```
metlos: This would prevent an update object that would only update properties. See line 72 and the use of `valueOrDefault()` method.

While metric type unit is currently the only "attribute" behaving that way, the resource type version used to behave like that in the past, too, and I think we should keep that. If a value of some attribute is not specified in the update object, the value from the original entity should be used.

Actually, I am not sure why the "properties" don't seem to behave that way, too. That is probably a bug, but that should be handled separately.

metlos: The above is still true. While you added the code to check for unit's nullity in the backend, in here you're essentially disallowing it ever be null. I think what you did for properties is correct, but unit is still not behaving correctly, IMHO.

pilhuhn: There are three points to this:
- granularity of updates (does the client need to supply all fields or are only the supplied ones updated)
- should we ever allow null as a valid value for MetricUnits
- are there fields like unit that are probably read-only (after first setting)? If a client can change them at will, what are the actual semantics of that? Will e.g. the metrics all update too? What would be the result of a change between incompatible units (e.g. time to kilobytes)

jkremser: > does the client need to supply all fields or are only the supplied ones updated

the latter, although there is an exception to this for the properties where we allow arbitrarily deep `Map<String, Object>` structure. If the null properties are passed in the update object, nothing is changed, empty hashmap sets it to empty properties; and non-empty hashmap sets (not updates only the keys that are set) it to the passed non-empty hashmap

> should we ever allow null as a valid value for MetricUnits

`null` in context of updating means "don't touch this"

> are there fields like unit that..

agree, it's tricky. I can think of an example where we/feed/whoever discover something and we don't know the units, so we set it to `NONE` and user can manually set the units to enrich the inventory metadata

pilhuhn: On 13 Jul 2015, at 17:53, Jirka Kremser wrote:

> agree, it's tricky. I can think of an example where we/feed/whoever 
> discover something and we don't know the units, so we set it to `NONE` 
> and user can manually set the units to enrich the inventory metadata

Yes. That would count as the first/initial setting.
But beyond that I believe that it does not necessarily make sense.

## 

hwr@pilhuhn.de  m:0179/207 4919  b:http://pilhuhn.blogspot.com

metlos:
lgtm, congrats to your first PR, Pavol! :smile:

PR #108, opened 2015-07-13
[HWKINVENT-88] Fix stable ID computation in REST

metlos:


api/src/main/java/org/hawkular/inventory/api/model/CanonicalPath.java
```diff
@@ -278,27 +278,43 @@ public String getEnvironmentId() {
         }
 
         public String getMetricTypeId() {
-            return idIfTypeCorrect(getRoot().down(), MetricType.class);
+            return idIfTypeCorrect(CanonicalPath.this, Resource.class);
```
jkremser: `s/Resource/Metric(Type)/` ?

metlos: Oh, the joys of copy paste :) Thanks for spotting it.

api/src/main/java/org/hawkular/inventory/api/model/CanonicalPath.java
```diff
@@ -278,27 +278,43 @@ public String getEnvironmentId() {
...
         }
 
         public String getResourceTypeId() {
-            return idIfTypeCorrect(getRoot().down(), ResourceType.class);
+            return idIfTypeCorrect(CanonicalPath.this, Resource.class);
```
jkremser: This looks suspicious too. `getResourceTypeId` -> `ResourceType.class` (?)

api/src/main/java/org/hawkular/inventory/api/model/CanonicalPath.java
```diff
@@ -278,27 +278,43 @@ public String getEnvironmentId() {
...
+            }
+
+            return ret.get();
         }
 
         public String getMetricId() {
-            String id = idIfTypeCorrect(getRoot().down(2), Metric.class);
-
-            return id != null ? id : idIfTypeCorrect(getRoot().down(3), Metric.class);
+            return idIfTypeCorrect(CanonicalPath.this, Resource.class);
```
jkremser: ---||---

PR #111, opened 2015-07-14
HWKINVENT-86 Adding gzip compression

jkremser:


metlos:
So does this mean GZIP is going to be used every time or is it negotiable between the client and the server?

jkremser:
Correct. If the `Accept-Encoding` header is set to `gzip, deflate` in the request like in here http://git.io/vmciB, then the server response should be gzipped (`Content-Encoding: gzip` header should be there and `Content-Length` should be lower)

Here is a curl that tests if it works: 

``` bash
curl -ivX GET -H "Accept: application/json" -H "Content-Type: application/json" -H "Accept-Encoding: gzip, deflate" 'http://jdoe:password@127.0.0.1:8080/hawkular/inventory/environments'
```

PR #113, opened 2015-07-15
[HWKINVENT-81] Adding more tests so that we're reasonably sure this works

metlos:
The querying was actually fixed by changes made for HWKINVENT-88 but this
test was never enabled within that unit of work.


metlos:
On the second thought, let me investigate this a little bit more ;) Don't merge yet pls.

metlos:
So one more change was needed, but it should be ready now.

ppalaga:
Perfect!

PR #119, opened 2015-07-20
org.jboss.logging.{LogMessage|MessageLogger|etc.} have been deprecate…

jkremser:
…d (just a change in the package name)


metlos:
whoops... too much dependence on autocomplete ;) Thanks for noticing.

PR #123, opened 2015-07-21
HWKINVENT-93 | Metric and Resource type under feed

pavolloffay:


api/src/main/java/org/hawkular/inventory/api/model/Path.java
```diff
@@ -834,17 +834,23 @@ private boolean fillPossiblePathsToTarget(Class<?> targetType, Class<?> currentT
 
             if (targetType.equals(currentType)) {
                 if (isStart) {
-                    result.add(currentType);
+                    if (!result.contains(currentType)) {
```
metlos: Is this necessary? `result` is a set, so a straight add should actually be cheaper than a check + add - both will try to look up if the element is present.

api/src/main/java/org/hawkular/inventory/base/BaseFeeds.java
```diff
@@ -160,5 +174,17 @@ public Multiple(TraversalContext<BE, Feed> context) {
         public Metrics.ReadContained metrics() {
             return new BaseMetrics.ReadContained<>(context.proceedTo(contains, Metric.class).get());
         }
+
+        @Override
+        public MetricTypes.ReadWrite metricTypes() {
+            // todo
```
metlos: ding, ding :)

api/src/main/java/org/hawkular/inventory/api/model/Path.java
pavolloffay: corrected

api/src/main/java/org/hawkular/inventory/base/BaseFeeds.java
pavolloffay: Ups... my big mistake.. Implemented to ReadContained.

PR #124, opened 2015-07-21
Unit tests for CanonicalPaths containing the escaped characters.

jkremser:
~~**tests only**~~

~~This will obviously fail the travis build~~

UPDATE: The implementation made by @metlos has landed to this PR

### DIFF
--- a/api/src/main/java/org/hawkular/inventory/api/model/Path.java
+++ b/api/src/main/java/org/hawkular/inventory/api/model/Path.java
@@ -477,6 +477,7 @@ public abstract class Path {
             //1 = reading id
             //2 = reading escape char
             int state = 0;
+            int stateBeforeEscapeChar = -1;
 
             loop:
             while (!progress.isFinished()) {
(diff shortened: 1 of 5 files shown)


metlos:
Ok, now there's the question of who of the two of us should merge this :smile:

metlos:
@ppalaga could you do an overall review and merge if ok, pls?

jkremser:
On the hawkular lvl, the issue is still there.. investigating..

jkremser:
merging myself as it's needed for hawkular master to work without errors and no one with push rights is currently online

jkremser:
@metlos here is the itest https://github.com/hawkular/hawkular/tree/dev/esc

I believe it corresponds to what hawkular-agent currently sends in here http://git.io/vYK60 I.e. in the json payload it's the basic `x.replaceAll("/", "\\\\/")` and in the URL it's `Util.urlEncode(x.replaceAll("/", "\\\\/"))`

jkremser:
It looks like we can have the better control of how the url is decoded by resteasy by using [UriInfo.getPath(false)](https://docs.oracle.com/javaee/7/api/javax/ws/rs/core/UriInfo.html#getPath-boolean-).

PR #126, opened 2015-07-23
HWKINVENT-97  | bus removed dependency on gson, bus test

pavolloffay:


bus-api/pom.xml
```diff
@@ -45,7 +45,7 @@
     <dependency>
       <groupId>org.hawkular.bus</groupId>
       <artifactId>hawkular-bus-common</artifactId>
-      <version>0.0.5</version>
+      <version>0.3.0.Final</version>
```
jmazzitelli: You should not be hardcoding component versions like this.  Extract out into a <version.org.hawkular.bus> property at least in the top parent pom.xml of your component (we should probably put these in parent pom in the future)

bus/pom.xml
```diff
@@ -49,40 +49,37 @@
...
       <scope>provided</scope>
     </dependency>
 
+    <!-- Test -->
     <dependency>
-      <groupId>org.jboss.logging</groupId>
-      <artifactId>jboss-logging</artifactId>
+      <groupId>org.hawkular.bus</groupId>
+      <artifactId>hawkular-bus-test-common</artifactId>
+      <version>0.3.0.Final</version>
```
jmazzitelli: You should not be hardcoding component versions like this.  Extract out into a <version.org.hawkular.bus> property at least in the top parent pom.xml of your component (we should probably put these in parent pom in the future)

pavolloffay: Thanks for comment :+1:

bus/src/test/java/org.hawkular.inventory.bus/BusTest.java
```diff
@@ -0,0 +1,162 @@
...
+    private static final Map<String, String> messageHeaders = new HashMap<>();
+
+    @BeforeClass
+    public static void init() {
+        objectProperties.put(PROP_KEY, PROP_VALUE);
+        messageHeaders.put(PROP_KEY, PROP_VALUE);
+    }
+
+    @Test
+    public void messagesSerializationTest() {
```
metlos, line 75: What is this testing?

metlos:
Apart from the test that IMHO doesn't test much and should be finished, this looks good.

bus/src/test/java/org.hawkular.inventory.bus/BusTest.java
jkremser, line 75: I was looking into this PR and the messages are serialized by the default object mapper. Pavol asked mazz if it'd be possible to configure own jackson object mapper for the bus-common. It should be available in the 0.3.1.Final.

Currently when the default JSON serialization is used the messages look ugly because of the path field, which uses the `toString()` representation. Also, I remember talking about it on IRC that it could worth trying to serialize the entity paths with the tenant ids, because the consumer on the bus can't figure out the tenant id.

However, this sub-task kind of complicates the original task which was only removing the gson dep., so imho we can merge and create another jira.

metlos, line 75: The test currently just creates instances of the event objects. As far as I can see it doesn't do anything with them. I assume that's because of the reasons you detailed but I'd have known if there was a TODO note in the tests, too ;)

I'm just paranoid about leaving half-finished stuff lying around because there's just lots of things on our plate and things will get forgotten about if not tracked in a JIRA or at least a TODO note.

metlos, line 75: In another words, if you point me to a JIRA or write a TODO note to the test, I'll merge straight away ;)

jmazzitelli:
See https://issues.jboss.org/browse/HAWKULAR-477 - it shows an example of what two methods you implement in your BasicMessage subclass in order to provide your own ObjectMapper.

pavolloffay:
I have updated the tests, and also configured object mapper. Now events posted on the bus contains full path with tenant id. 

I found one bug related to inventory bus https://issues.jboss.org/browse/HWKINVENT-98. Currently the is error when intventory tries to connect to the bus.

metlos:
Perfect work Pavol!

PR #127, opened 2015-07-24
HWKINVENT-101 - Shutting down hawkular server takes ages

jkremser:
What dramatically\* shortens the shutdown time is this line

```
storage.backend=cassandrathrift
```

Other options are there for Titan fine tuning.

\* from 1-2 minutes into 1-2 seconds!

### DIFF
--- a/hawkular-integrated-inventory-rest/src/main/resources/hawkular-inventory.properties
+++ b/hawkular-integrated-inventory-rest/src/main/resources/hawkular-inventory.properties
@@ -15,7 +15,9 @@
 # limitations under the License.
 #
 
-storage.backend=cassandra
+# see this documentation for more options http://s3.thinkaurelius.com/docs/titan/0.5.4/titan-config-ref.html
+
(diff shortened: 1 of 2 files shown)


jkremser:
I marked this PR with `don't merge yet` label, because when I added an URL, kill the server immediately after it, and start the server again the URL wasn't there.

After further investigation, however, there is no difference between status quo (without this change) and this PR, the URL was missing even with the old 1-2 minutes shutdown behavior. So I am removing the flag.

metlos:
Were you able to get to the bottom of the kill-killing-commit issue? If not yet, could you create a JIRA for it so that it doesn't get forgotten about?

I'll merge this now, but I would like to follow up on the disappearing commits problem.

jkremser:
> Were you able to get to the bottom of the kill-killing-commit issue? If not yet, could you create a JIRA for it so that it doesn't get forgotten about?

HWKINVENT-103

PR #128, opened 2015-07-27
HWKINVENT 102 - various logging stuff including other miscellaneous things ;)

jkremser:
It actually contains 3 small things:
- Using the jackson version from parent pom + replacing the deprecated ObjectMapper.addMixInAnnotations with addMixIn
- Mapping the *NotFound and *AlreadyExists exceptions to debug log level (no warnings anymore)
- logging interceptor for all rest HTTP requests if DEBUG or lower

### DIFF
--- a/rest-servlet/src/main/java/org/hawkular/inventory/rest/interceptors/LoggingInterceptor.java
+++ b/rest-servlet/src/main/java/org/hawkular/inventory/rest/interceptors/LoggingInterceptor.java
@@ -27,11 +27,10 @@ import javax.ws.rs.core.MultivaluedMap;
 import javax.ws.rs.ext.Provider;
 
 import com.fasterxml.jackson.databind.ObjectMapper;
-import com.fasterxml.jackson.databind.ObjectWriter;
-import com.fasterxml.jackson.databind.SerializationFeature;
 
 import org.apache.commons.io.IOUtils;
 import org.hawkular.inventory.rest.RestApiLogger;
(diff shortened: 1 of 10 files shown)


rest-servlet/src/main/java/org/hawkular/inventory/rest/interceptors/LoggingInterceptor.java
```diff
@@ -0,0 +1,74 @@
...
+public class LoggingInterceptor implements ContainerRequestFilter {
+
+    private static final ObjectMapper MAPPER = new ObjectMapper();
+    static {
+        JacksonConfig.initializeObjectMapper(MAPPER);
+    }
+
+    @Override
+    public void filter(ContainerRequestContext containerRequestContext) throws IOException {
+        // perhaps better than logger lvl could be a system property passed by -Dx=y
```
metlos: You could also reserve a special logger category for this so that users can switch this on/off independently of debug logging of the rest of the REST api - should be as easy as adding a new constant to the `RestApiLogger` interface and using it here.

jkremser: It should be done now.

metlos:
checkstyle failure

jkremser:
> checkstyle failure

ouch
should be ok, now

metlos:
Great, merging!

PR #131, opened 2015-07-31
HWKINVENT-107 Prefix artifactIds with hawkular- and rename module directories 

ppalaga:


ppalaga:
@metlos and @Jiri-Kremser could you please review?

ppalaga:
All levels should be fixed in 9b8864a

pom.xml
```diff
@@ -44,18 +56,6 @@
...
 
-  <modules>
-    <module>api</module>
-    <module>hawkular-inventory-json-helper</module>
-    <module>impl-tinkerpop-parent</module>
-    <module>hawkular-inventory-cdi</module>
-    <module>rest-servlet</module>
-    <module>bus</module>
-    <module>bus-api</module>
-    <module>hawkular-integrated-inventory-rest</module>
```
jkremser, line 55: after we release the change will be needed also on `hawkular/hawkular` side (to depend on `hawkular-inventory-dist` instead of `hawkular-integrated-inventory-rest`)

ppalaga, line 55: Yes, that's expected. Changes will perhaps also be needed in Agent. IMO this is a reason to release inventory ASAP after merging this and change the dep names in downstream.

ppalaga:
Resolved conflicts with master

hawkular-inventory-api/pom.xml
```diff
@@ -78,7 +79,6 @@
 
   <build>
     <finalName>hawkular-${project.artifactId}-${project.version}</finalName>
```
metlos, line 81: There's no need for this anymore as this should be the same as the default name now.

metlos:
K, I'm flipping the switch :) Thanks for this, @ppalaga

PR #134, opened 2015-07-31
HWKINVENT-109 | BUS bugfix sending messages

pavolloffay:


hawkular-inventory-bus/src/main/java/org/hawkular/inventory/bus/BusIntegration.java
```diff
@@ -61,21 +56,12 @@ public void configure(Configuration configuration) {
         this.configuration = configuration;
     }
 
-    public void start() throws NamingException, JMSException {
+    public void start() {
         if (namingContext != null) {
```
metlos, line 63: As far as I can see, `namingContext` is never assigned, so this check is pointless.

metlos, line 63: Also the `stop()` method will throw an NPE on trying to close the `namingContext`.

hawkular-inventory-bus/src/main/java/org/hawkular/inventory/bus/MessageSender.java
```diff
@@ -32,23 +37,26 @@
...
+
+    public MessageSender(Configuration configuration) {
+        this.configuration = configuration;
         this.messageProcessor = new MessageProcessor();
     }
 
     public void send(Interest<?, ?> interest, Object inventoryEvent) {
         InventoryEvent<?> message = InventoryEvent.from(interest.getAction(), inventoryEvent);
         try {
+            ProducerConnectionContext producerConnectionContext = init();
```
metlos: Doesn't this leak connections or something? I'm not sure, just asking if re-initializing everything for each message sent is really what we want here.

hawkular-inventory-bus/src/main/java/org/hawkular/inventory/bus/MessageSender.java
```diff
@@ -65,4 +73,18 @@ public void send(Interest<?, ?> interest, Object inventoryEvent) {
     private String firstLetterLowercased(String source) {
         return Character.toLowerCase(source.charAt(0)) + source.substring(1);
     }
+
+    private ProducerConnectionContext init() throws JMSException, NamingException {
+        InitialContext namingContext = new InitialContext();
```
metlos: `InitialContext` should be `close()`d afer usage, because it can leak resources.

pavolloffay: I wanted to comment don't merge yet. But you were so fat.  Yes it's not efficient to reinitialize it every time message is sent... 

I will correct when I get home or in the morning.

jkremser:
lgtm,
thanks!

PR #135, opened 2015-08-03
Order imports to comply with the new Checkstyle rule

ppalaga:


ppalaga:
Will be replaced by a new PR

PR #136, opened 2015-08-03
GraphSON dump + get 1 relationship info

jkremser:
- The first commit brings the support for the whole graph dump.
- The second one fixes the HWKINVENT-110

TODO:
- [x] make use of the OutputStream instead of String for the graph representation

### DIFF
--- a/hawkular-inventory-api/src/main/java/org/hawkular/inventory/api/EmptyInventory.java
+++ b/hawkular-inventory-api/src/main/java/org/hawkular/inventory/api/EmptyInventory.java
@@ -73,6 +73,11 @@ public class EmptyInventory implements Inventory {
         return Observable.empty();
     }
 
+    @Override
+    public String getGraphSON(String tenantId) {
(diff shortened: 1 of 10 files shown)


metlos:
Idea and impl looks mostly good. The only thing I would like to investigate is to replace storing the whole graphson in a string with outputting the graphson data directly to an outputstream to limit the memory usage.

The graphs will tend to grow big.

If this would prove too difficult, we can merge this like it is now and create a JIRA to replace it with a streaming impl later. But I want to do an informed decision on that.

jkremser:
> If this would prove too difficult, we can merge this like it is now and create a JIRA to replace it with a streaming impl later. But I want to do an informed decision on that.

nope, it's fine, I just forgot, it's on my todo list :]

jkremser:
thanks, here is the subsequent work captured by the JIRA https://issues.jboss.org/browse/HWKINVENT-111

PR #137, opened 2015-08-05
Make java imports comply to the new Checkstyle rule and IDE settings

ppalaga:


ppalaga:
@metlos could you please review and merge. Please note that this PR is quite big and has high chances to conflict with the subsequent work.

PR #138, opened 2015-08-06
[HWKINVENT-62] Structured data data model, access and serialization.

metlos:
This has some rough edges still, but is complete enough to publicize so that peer review can start.
- [x] - more unit tests
- [x] - `hasData` should not be user-creatable
- [x] - `DataEntity` json serialization
- [x] - Reword and add javadoc
- [x] - Clean up (some unimplemented methods that don't need implementing anymore, etc.)

### DIFF
--- a/hawkular-inventory-api/src/main/java/org/hawkular/inventory/api/Datas.java
+++ b/hawkular-inventory-api/src/main/java/org/hawkular/inventory/api/Datas.java
@@ -79,7 +79,7 @@ public final class Datas {
          *
          * @param dataPath the path to the subset of the data.
          * @return the subset of the data stored with the data entity
-         * @see #bareData(RelativePath)
+         * @see #flatData(RelativePath)
          */
         StructuredData data(RelativePath dataPath);
 
@@ -93,12 +93,12 @@ public final class Datas {
          * @param dataPath the path to the subset of the data to return
          * @return the subset of the data stored with the data entity
          */
-        StructuredData bareData(RelativePath dataPath);
+        StructuredData flatData(RelativePath dataPath);
 
         @Override
         default boolean exists() {
             try {
-                bareData(RelativePath.empty().get());
+                flatData(RelativePath.empty().get());
                 return true;
             } catch (EntityNotFoundException | RelationNotFoundException ignored) {
                 return false;
@@ -128,17 +128,17 @@ public final class Datas {
         Page<StructuredData> datas(RelativePath dataPath, Pager pager);
 
         /**
-         * Similar to {@link Single#bareData(RelativePath)}, only resolved over multiple entities.
+         * Similar to {@link Single#flatData(RelativePath)}, only resolved over multiple entities.
          *
          * @param dataPath the path to the data entry inside the data entity
          * @param pager    pager to use to page the data
          * @return the page of the data entries
          */
-        Page<StructuredData> bareDatas(RelativePath dataPath, Pager pager);
+        Page<StructuredData> flatDatas(RelativePath dataPath, Pager pager);
 
         @Override
         default boolean anyExists() {
-            return !bareDatas(RelativePath.empty().get(),
+            return !flatDatas(RelativePath.empty().get(),
                     Pager.builder().withPageSize(1).orderBy(Order.unspecified()).build()).isEmpty();
         }
     }
--- a/hawkular-inventory-api/src/main/java/org/hawkular/inventory/api/Datas.java
+++ b/hawkular-inventory-api/src/main/java/org/hawkular/inventory/api/Datas.java
@@ -1,0 +1,145 @@
+/*
+ * Copyright 2015 Red Hat, Inc. and/or its affiliates
+ * and other contributors as indicated by the @author tags.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.hawkular.inventory.api;
+
+import org.hawkular.inventory.api.model.DataEntity;
+import org.hawkular.inventory.api.model.RelativePath;
+import org.hawkular.inventory.api.model.StructuredData;
+import org.hawkular.inventory.api.paging.Order;
+import org.hawkular.inventory.api.paging.Page;
+import org.hawkular.inventory.api.paging.Pager;
+
+/**
+ * @author Lukas Krejci
+ * @since 0.3.0
+ */
+public final class Datas {
+
+    private Datas() {
+
+    }
+
+    protected interface ReadBase {
+
+        /**
+         * @return the access to the resource this data belongs to.
+         */
+        Resources.Single resource();
+    }
+
+    public interface Read extends ReadInterface<Single, Multiple, Void>, ReadBase {
+    }
+
+    public interface ReadWrite
+            extends ReadWriteInterface<DataEntity.Update, DataEntity.Blueprint, Single, Multiple, Void>,
+            ReadBase {
+    }
+
+    public interface Single extends ResolvableToSingle<DataEntity, DataEntity.Update> {
+
+        /**
+         * Loads the data entity on the current position in the inventory traversal along with its data.
+         *
+         * <p>Note that this might be a potentially expensive operation because of the attached data structure being
+         * loaded.
+         *
+         * @return the fully loaded structured data on the current position in the inventory traversal
+         * @throws EntityNotFoundException if there is no structured data on the current position in the inventory
+         * @see ResolvableToSingle#entity()
+         */
+        @Override
+        DataEntity entity() throws EntityNotFoundException;
+
+        /**
+         * Returns the data on the path relative to the entity.
+         *
+         * In another words, you can use this method to obtain only a subset of the data stored on the data entity.
+         *
+         * <p>I.e if you have an data entity which contains a map with a key "foo", which contains a list and you
+         * want to obtain a third element of that list, you'd do:
+         * {@code
+         * ...data(RelativePath.to().structuredData().key("foo").index(2).get());
+         * }
+         *
+         * <p>If you want to obtain the whole data structure, use an empty path: {@code RelativePath.empty().get()}.
+         *
+         * @param dataPath the path to the subset of the data.
+         * @return the subset of the data stored with the data entity
+         * @see #bareData(RelativePath)
+         */
+        StructuredData data(RelativePath dataPath);
+
+        /**
+         * This is very similar to {@link #data(RelativePath)} but this method doesn't load the child data.
+         *
+         * <p>If the data on the path contains a "primitive" value, the value is loaded. If the data contains a list,
+         * the returned instance will contain an empty list and if the data contains a map the returned instance will
+         * contain an empty map.
+         *
+         * @param dataPath the path to the subset of the data to return
+         * @return the subset of the data stored with the data entity
+         */
+        StructuredData bareData(RelativePath dataPath);
+
+        @Override
+        default boolean exists() {
+            try {
+                bareData(RelativePath.empty().get());
+                return true;
+            } catch (EntityNotFoundException | RelationNotFoundException ignored) {
+                return false;
+            }
+        }
+    }
+
+    public interface Multiple extends ResolvableToMany<DataEntity> {
+
+        /**
+         * Note that this is potentially expensive operation because it loads all the data associated with each of the
+         * returned data entities.
+         *
+         * @param pager the pager object describing the subset of the entities to return
+         * @return the page of the results
+         */
+        @Override
+        Page<DataEntity> entities(Pager pager);
+
+        /**
+         * Similar to {@link Single#data(RelativePath)}, only resolved over multiple entities.
+         *
+         * @param dataPath the path to the data entry inside the data entity
+         * @param pager    pager to use to page the data
+         * @return the page of the data entries
+         */
+        Page<StructuredData> datas(RelativePath dataPath, Pager pager);
+
+        /**
+         * Similar to {@link Single#bareData(RelativePath)}, only resolved over multiple entities.
+         *
+         * @param dataPath the path to the data entry inside the data entity
+         * @param pager    pager to use to page the data
+         * @return the page of the data entries
+         */
+        Page<StructuredData> bareDatas(RelativePath dataPath, Pager pager);
+
+        @Override
+        default boolean anyExists() {
+            return !bareDatas(RelativePath.empty().get(),
+                    Pager.builder().withPageSize(1).orderBy(Order.unspecified()).build()).isEmpty();
+        }
+    }
+}
--- a/hawkular-inventory-api/src/main/java/org/hawkular/inventory/api/EmptyInventory.java
+++ b/hawkular-inventory-api/src/main/java/org/hawkular/inventory/api/EmptyInventory.java
@@ -1248,7 +1248,7 @@ public class EmptyInventory implements Inventory {
 
     public static class DatasSingle implements Datas.Single {
         @Override
-        public StructuredData bareData(RelativePath dataPath) {
+        public StructuredData flatData(RelativePath dataPath) {
             throw entityNotFound(null);
         }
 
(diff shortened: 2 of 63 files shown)


hawkular-inventory-api/src/main/java/org/hawkular/inventory/api/Datas.java
```diff
@@ -0,0 +1,145 @@
...
+         * This is very similar to {@link #data(RelativePath)} but this method doesn't load the child data.
+         *
+         * <p>If the data on the path contains a "primitive" value, the value is loaded. If the data contains a list,
+         * the returned instance will contain an empty list and if the data contains a map the returned instance will
+         * contain an empty map.
+         *
+         * @param dataPath the path to the subset of the data to return
+         * @return the subset of the data stored with the data entity
+         */
+        StructuredData bareData(RelativePath dataPath);
```
jkremser: This is here because of performance reasons, right? Perhaps the `flatData()` would be more descriptive name here.

EDIT:
If I understand it well, you can still go into deep by the relative path, but it wont fetch the data recursively from that point deeper, is that correct?

so 

``` java
...bareData(RelativePath.to().structuredData().key("foo").index(2).get());
```

does the same as `...data(same_param)` except it returns only the flat (primitive values) on that position (?)

metlos: Exactly. And I'll rename it to `flatData` - good suggestion.

hawkular-inventory-api/src/main/java/org/hawkular/inventory/api/Datas.java
```diff
@@ -0,0 +1,145 @@
...
+        Page<DataEntity> entities(Pager pager);
+
+        /**
+         * Similar to {@link Single#data(RelativePath)}, only resolved over multiple entities.
+         *
+         * @param dataPath the path to the data entry inside the data entity
+         * @param pager    pager to use to page the data
+         * @return the page of the data entries
+         */
+        Page<StructuredData> datas(RelativePath dataPath, Pager pager);
```
jkremser, line 128:
```
<grammar_nazi>
"data" in plural and "datum" is singular
</grammar_nazi>
```

Although, I saw using the term 'datas' and I am not native speaker, so I don't really care :]

metlos, line 128: Well, config as such is not a single "datum" - it's data already ;) I don't like this myself either, but hope it gets the point across.

hawkular-inventory-api/src/main/java/org/hawkular/inventory/api/Relationships.java
```diff
@@ -87,7 +87,30 @@ private Relationships() {
          * <p>When deleting a resource, all its contained child resources are deleted along with it as mandated by the
          * contract of the {@link #contains} relationship.
          */
-        isParentOf
+        isParentOf,
+
+        /**
+         * A relationship from a resource to a structural data that contains the configuration of that resource.
+         *
+         * <p>The relationship is 1-to-1. There is no sharing of configuration objects between different resources.
```
jkremser, line 95: `0-1 to 0-1` right? Configuration can live without a resource and vice-versa. I can imagine that there will be a lot of leaves in the graph (resource nodes) without any configuration and the more abstract ones (higher in the resrouce hierarchy) will have the configuration. Perhaps some kind of following of the resource hierarchy could make a sense in the future + the ability to override it in the children.

metlos, line 95: right, resource to config is 1 to 0..1. Right now a config cannot live without a resource - I understand config as just data, so it only makes sense when connected to some entity. Also I explicitly don't allow more resources reference a single config object. Even though it would make sense from the point of view of lower storage reqs, I don't think it would be a good idea "semantically" - each resource is a standalone entity and is treated separately in the feeds. So it would I think not be clear what to do if several resources shared a config object and suddenly a feed would like to update config of one of them.

hawkular-inventory-api/src/main/java/org/hawkular/inventory/api/filters/With.java
```diff
@@ -119,6 +122,39 @@ public static PropertyValues propertyValues(String name, Object... values) {
         return new PropertyValues(name, values);
     }
 
+    /**
+     * Checks whether there is data on the provided relative path. This filter is only really applicable
+     * on the {@link org.hawkular.inventory.api.model.DataEntity}. I.e. the relative path is always resolved
```
jkremser, line 127: > This filter is only really applicable on the {@link org.hawkular.inventory.api.model.DataEntity}

So it's similar as the `RelationWith` thing, right (i.e. not completely type safe)? What about having `WithData` class?

hawkular-inventory-api/src/main/java/org/hawkular/inventory/api/model/DataEntity.java
```diff
@@ -0,0 +1,165 @@
...
+    }
+
+    @Override
+    public Updater<Update, ? extends AbstractElement<?, Update>> update() {
+        return new Updater<>((u) -> new DataEntity(this.getPath().up(), this.getRole(),
+                valueOrDefault(u.getValue(), getValue()), u.getProperties()));
+    }
+
+    public enum Role {
+        configuration, connectionConfiguration;
```
jkremser, line 74: I know operations are not in scope of this PR, but do you plan to represent it by the DataEntity in the future?

hawkular-inventory-api/src/main/java/org/hawkular/inventory/api/model/DataEntity.java
```diff
@@ -0,0 +1,165 @@
...
+            }
+
+            @Override
+            public Blueprint build() {
+                return new Blueprint(value, properties);
+            }
+        }
+    }
+
+    public static final class Update extends Entity.Update {
```
jkremser, line 129: I guess the `role` is not here and in the Blueprint.Builder, because it's inferred from the context of the api call, right? If called on `...configuration()` or `...connectConfig()`

hawkular-inventory-api/src/main/java/org/hawkular/inventory/api/model/IdentifiedEntity.java
```diff
@@ -0,0 +1,73 @@
...
+        super(path, properties);
+        if (!this.getClass().equals(path.getSegment().getElementType())) {
+            throw new IllegalArgumentException("Invalid path specified. Trying to create " +
+                    this.getClass().getSimpleName() + " but the path points to " +
+                    path.getSegment().getElementType().getSimpleName());
+        }
+    }
+
+    public abstract static class Blueprint extends AbstractElement.Blueprint {
+        @XmlAttribute
```
jkremser, line 50: Is the `@XmlAttribute` needed? I know we have this also in `AbstractElement` and other places, but why? Does jackson use that? I don't think so. Do we serialize the classes into xml somewhere?

jkremser:
What about the config type - config instance thingy? From the code, it entails that each resource can have different config, so no types. I am good with it, but I wonder how for instance we display some kind of general config editor if the resource configuration contains only the set values.

Perhaps it's doable by the `Type.undefined`, but it's kind of cumbersome, imho.

hawkular-inventory-api/src/main/java/org/hawkular/inventory/base/BaseTenants.java
```diff
@@ -46,7 +46,7 @@ private BaseTenants() {
 
     }
 
-    public static class ReadWrite<BE> extends Mutator<BE, Tenant, Tenant.Blueprint, Tenant.Update>
+    public static class ReadWrite<BE> extends Mutator<BE, Tenant, Tenant.Blueprint, Tenant.Update, String>
```
jkremser, line 49: not sure I like the string "String" as a type parameter, you basically override/shadow the basic java type with it. What if I wanted to add simple helper method which will use java.lang.String? I'd have to use the fully qualified path, this may be confusing, ID was imho better.

hawkular-inventory-api/src/main/java/org/hawkular/inventory/api/model/DataEntity.java
metlos, line 74: Possibly. But those would be operation results or passed in parameters. The operation definitions would be akin to config definitions and those are not part of this PR either and I think would look different - something like json schema or such like.

hawkular-inventory-api/src/main/java/org/hawkular/inventory/api/model/DataEntity.java
metlos, line 129: Exactly.

hawkular-inventory-api/src/main/java/org/hawkular/inventory/api/model/IdentifiedEntity.java
metlos, line 50: No we don't need that. I just refactored this out of the original `Entity` (for a reason I don't actually recall) so it just stayed where it was.

metlos:
> What about the config type - config instance thingy? From the code, it entails that each resource can have different config, so no types. I am good with it, but I wonder how for instance we display some kind of general config editor if the resource configuration contains only the set values.
> 
> Perhaps it's doable by the Type.undefined, but it's kind of cumbersome, imho.

config defs will be a subsequent PR. I think we're going to end up with a json schema for those. But I thought having the ability to store and search for config is currently more important than having the ability to define metadata about it.

hawkular-inventory-api/src/main/java/org/hawkular/inventory/base/BaseTenants.java
metlos, line 49: That's the actual `java.lang.String`. `Mutator` has an `Id` type parameter and `ReadWrite` extends mutator with the `Id` type param "realized" with `String`.

metlos, line 49: The reason for adding this is that configs don't have any IDs so they use `Void` as their id type. The configs are identified by the role. Which brings me to a possible refactoring:

Instead of `configuration` and `connectionConfiguration` methods in `Resources.Single|Multiple` we could use the `DataEntity.Role` as the id and have:
`Resources.Single#data(DataEntity.Role)` method. This would IMHO be maybe better than the current approach of a method per data role.

hawkular-inventory-api/src/test/java/org/hawkular/inventory/api/test/AbstractBaseInventoryPersistenceCheck.java
```diff
@@ -248,6 +258,69 @@ private void setupData() throws Exception {
...
+                .putMap("mapOfMaps")
+                /**/.putMap("map1")
+                /**//**/.putIntegral("int", 42L)
+                /**//**/.putString("answer", "probably")
+                /**/.closeMap()
+                /**/.putMap("map2")
+                /**//**/.putFloatingPoint("float", 2.0)
+                /**/.closeMap()
+                .closeMap()
+                .build();
```
jkremser, line 318: nice, uber builder!

hawkular-inventory-api/src/main/java/org/hawkular/inventory/api/filters/With.java
metlos, line 127: What would that class do?

hawkular-inventory-api/src/main/java/org/hawkular/inventory/api/EmptyInventory.java
```diff
@@ -1170,4 +1194,99 @@ public Relationship associationWith(Path path) throws RelationNotFoundException
...
+            return new ResourcesSingle();
+        }
+
+        @Override
+        public Datas.Multiple getAll(Filter[][] filters) {
+            return new DatasMultiple();
+        }
+
+        @Override
+        public Datas.Single get(Void ignored) throws EntityNotFoundException {
```
jkremser, line 1229: pitta, Java hasn't support for the default parameters... the `.get(null)` in the API looks ugly

hawkular-inventory-api/src/test/java/org/hawkular/inventory/api/test/StructuredDataTest.java
```diff
@@ -0,0 +1,222 @@
...
+        assertEquals(1.0, floatingPoint.getValue());
+        assertEquals("kachny", string.getValue());
+        assertEquals(Boolean.TRUE, ((List<StructuredData>) list.getValue()).get(0).getValue());
+        assertEquals(1L, ((List<StructuredData>) list.getValue()).get(1).getValue());
+        assertEquals(Boolean.TRUE, ((Map<String, StructuredData>) map.getValue()).get("bool").getValue());
+        assertEquals(1L, ((Map<String, StructuredData>) map.getValue()).get("int").getValue());
+
+        // yes, this is why visitors are better ;)
+        assertEquals(Boolean.TRUE,
+                (((List<StructuredData>) (((List<StructuredData>) listInList.getValue()).get(0).getValue())).get(0)
```
jkremser, line 67: is this lisp? :]

hawkular-inventory-impl-tinkerpop-parent/hawkular-inventory-impl-tinkerpop/src/main/java/org/hawkular/inventory/impl/tinkerpop/FilterVisitor.java
```diff
@@ -302,4 +358,30 @@ private void apply(Path.Segment segment, HawkularPipeline<?, ?> pipeline) {
...
+
+    /**
+     * A very simplistic conversion of string to positive integer in only decimal radix.
+     *
+     * <p>This is used to figure out whether a segment id represents an index or a key.
+     *
+     * @param str the string potentially representing a number
+     * @return the parsed number or null if the string is not a supported number
+     */
+    private static Integer toInteger(String str) {
```
jkremser, line 370: why not Integer.parseInt?

hawkular-inventory-api/src/main/java/org/hawkular/inventory/api/filters/With.java
jkremser, line 127: well, basically the same thing as `RelationWith` does, i.e. narrowing the allowable filters only for the DataEntities. It's not typesafe, you can pass the `RelationWith.foo()` instance where normal filter is expected, but at least the api should look odd :]

```
inventory.tenants.getAll(RelationWith.name(foo)) X inventory.tenants.relationships.getAll(RelationWith.name(foo))
```

 Currently the data filters make sense only for resources, but it's on the `With` class that any entity filter can use.

hawkular-inventory-api/src/main/java/org/hawkular/inventory/base/BaseTenants.java
jkremser, line 49: > That's the actual java.lang.String

oh, ok it's confusing. So if the type parameter is has the same name as imported class or a class from `java.lang.` (which are implicitly imported), it's not type parameter but a concrete class..

> Resources.Single#data(DataEntity.Role) method. This would IMHO be maybe better than the current approach of a method per data role. 

yeah, Resources.Single#data(DataEntity.Role) sounds more extensible

hawkular-inventory-api/src/test/java/org/hawkular/inventory/api/test/StructuredDataTest.java
metlos, line 67: written that way so that comment on line 65 really strikes home :wink:

hawkular-inventory-impl-tinkerpop-parent/hawkular-inventory-impl-tinkerpop/src/main/java/org/hawkular/inventory/impl/tinkerpop/FilterVisitor.java
metlos, line 370: so that `try...catch` is not used for control flow. Basically it is expected to be the common case that the `str` is NOT a number. Because this is used inside query translation and hence needs to be fast, IMHO, I opted for a custom solution that will not throw exceptions on failure to parse.

jkremser, line 370: makes sense

hawkular-inventory-impl-tinkerpop-parent/hawkular-inventory-impl-tinkerpop/src/main/java/org/hawkular/inventory/impl/tinkerpop/FilterVisitor.java
```diff
@@ -302,4 +358,30 @@ private void apply(Path.Segment segment, HawkularPipeline<?, ?> pipeline) {
...
+        char[] chars = str.toCharArray();
```
metlos, line 371: This is not necessary and wasteful - `String.charAt()` should be used instead.

PR #140, opened 2015-08-11
HWKINVENT-103 | removed tinkergraph from classpath

pavolloffay:
Removed dependency on tinkerGraph from runtime. 

See OfficialInventoryProducer.java#102  I think this was also causing long startup of inventory.


metlos:
Great! Thanks for fixing these 2 hiccups of mine ;)

PR #141, opened 2015-08-11
HWKINVENT-109 | BUS - new connection is created every time

pavolloffay:
see  https://developer.jboss.org/wiki/ShouldICacheJMSConnectionsAndJMSSessions

The reason is, when you use the JMS API from inside an EJB (including an MDB), or a servlet, (or a MBean in JBoss), then when you call createConnection() or createSession() on your JMS connection factory or JMS connection, you're not actually making those calls on a real JMS connection factory or JMS connection.

The JEE specification mandates that the connection factory you use is actually a JCA managed connection factory (see the JCA specification). The JCA managed connection factory wraps the real underlying JMS connection factory and implements the same interface so to you, the user, you're just using a normal connection factory, right?

### DIFF
--- a/hawkular-inventory-bus/src/main/java/org/hawkular/inventory/bus/MessageSender.java
+++ b/hawkular-inventory-bus/src/main/java/org/hawkular/inventory/bus/MessageSender.java
@@ -22,6 +22,7 @@ import java.util.HashMap;
 import java.util.Map;
 
 import javax.jms.JMSException;
+import javax.jms.TopicConnectionFactory;
 
 import org.hawkular.bus.common.ConnectionContextFactory;
 import org.hawkular.bus.common.Endpoint;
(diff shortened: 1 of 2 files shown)


hawkular-inventory-bus/src/main/java/org/hawkular/inventory/bus/MessageSender.java
```diff
@@ -35,24 +36,25 @@
...
  */
 final class MessageSender {
-    private final MessageProcessor messageProcessor;
-    private final ConnectionContextFactory connectionContextFactory;
     private final String topicName;
-    private ProducerConnectionContext producerConnectionContext;
+    private final TopicConnectionFactory topicConnectionFactory;
+    private final MessageProcessor messageProcessor;
 
+    public MessageSender(TopicConnectionFactory  topicConnectionFactory, String topicName) {
```
jmazzitelli, line 43: are you always going to be sending to a Topic? If you are never going to use queues, this is OK, but if you would ever want to send to a queue, you'd need to pass in ConnectionFactory not one of [Topic,Queue]ConnectionFactory

pavolloffay, line 43: For the moment I think we are going to send them only to a topic

metlos, line 43: I agree with @pavolloffay. We can change this if and when the need for sending to a queue arises.

PR #142, opened 2015-08-11
HWKINVENT-113 - REST endpoint for feeds to use to create resource configurations

jkremser:
This depends on #144
~~It's not finished~~
I wanted to add the tests first, but they should go to another repo anyway.. when testing with curl, it works :]

### DIFF
--- a/hawkular-inventory-api/src/main/java/org/hawkular/inventory/api/model/DataEntity.java
+++ b/hawkular-inventory-api/src/main/java/org/hawkular/inventory/api/model/DataEntity.java
@@ -86,6 +86,11 @@ public class DataEntity extends Entity<DataEntity.Blueprint, DataEntity.Update> 
             this.value = value;
         }
 
+        // this is needed for Jackson deserialization
+        public Blueprint() {
+            this(null, null, null);
+        }
+
         public StructuredData getValue() {
             return value;
         }
--- a/hawkular-inventory-api/src/main/java/org/hawkular/inventory/api/model/DataEntity.java
+++ b/hawkular-inventory-api/src/main/java/org/hawkular/inventory/api/model/DataEntity.java
@@ -87,7 +87,7 @@ public class DataEntity extends Entity<DataEntity.Blueprint, DataEntity.Update> 
         }
 
         // this is needed for Jackson deserialization
-        public Blueprint() {
+        private Blueprint() {
             this(null, null, null);
         }
 
@@ -143,7 +143,7 @@ public class DataEntity extends Entity<DataEntity.Blueprint, DataEntity.Update> 
         }
 
         // this is needed for Jackson deserialization
-        public Update() {
+        private Update() {
             this(null, null);
         }
 
--- a/hawkular-inventory-api/src/main/java/org/hawkular/inventory/api/model/DataEntity.java
+++ b/hawkular-inventory-api/src/main/java/org/hawkular/inventory/api/model/DataEntity.java
@@ -130,6 +130,7 @@ public class DataEntity extends Entity<DataEntity.Blueprint, DataEntity.Update> 
     }
 
     public static final class Update extends Entity.Update {
+
         private final StructuredData value;
 
         public static Builder builder() {
@@ -139,6 +140,11 @@ public class DataEntity extends Entity<DataEntity.Blueprint, DataEntity.Update> 
         public Update(StructuredData value, Map<String, Object> properties) {
             super(properties);
             this.value = value;
+        }
+
+        // this is needed for Jackson deserialization
+        public Update() {
+            this(null, null);
         }
 
         public StructuredData getValue() {
--- a/hawkular-inventory-json-helper/src/main/java/org/hawkular/inventory/json/mixins/DataEntityMixin.java
+++ b/hawkular-inventory-json-helper/src/main/java/org/hawkular/inventory/json/mixins/DataEntityMixin.java
@@ -19,20 +19,26 @@ package org.hawkular.inventory.json.mixins;
 import java.util.Map;
 
 import org.hawkular.inventory.api.model.CanonicalPath;
+import org.hawkular.inventory.api.model.DataEntity;
 import org.hawkular.inventory.api.model.StructuredData;
 
 import com.fasterxml.jackson.annotation.JsonCreator;
+import com.fasterxml.jackson.annotation.JsonIgnore;
 import com.fasterxml.jackson.annotation.JsonProperty;
 
 /**
  * @author Lukas Krejci
  * @since 0.3.0
  */
-public class DataEntityMixin {
+public abstract class DataEntityMixin {
 
     @JsonCreator
     public DataEntityMixin(@JsonProperty("path") CanonicalPath path, @JsonProperty("value") StructuredData value,
             @JsonProperty("properties") Map<String, Object> properties) {
     }
+
+    @JsonIgnore abstract String getId();
+
+    @JsonIgnore abstract DataEntity.Role getRole();
 
 }
--- a/hawkular-inventory-rest-api/src/main/java/org/hawkular/inventory/rest/RestResources.java
+++ b/hawkular-inventory-rest-api/src/main/java/org/hawkular/inventory/rest/RestResources.java
@@ -567,7 +567,7 @@ public class RestResources extends RestBase {
     @javax.ws.rs.Path("/{environmentId}/{feedId}/resources/{resourcePath:.+}/data")
     @ApiOperation("Creates the configuration for pre-existing resource")
     @ApiResponses({
-                          @ApiResponse(code = 204, message = "OK"),
+                          @ApiResponse(code = 204, message = "OK Created"),
                           @ApiResponse(code = 404, message = "Tenant, environment, resource or feed doesn't exist",
                                        response = ApiError.class),
                           @ApiResponse(code = 500, message = "Server error", response = ApiError.class)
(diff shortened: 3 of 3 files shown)


hawkular-inventory-rest-api/src/main/java/org/hawkular/inventory/rest/RestResources.java
```diff
@@ -563,6 +564,117 @@ public Response associateMetrics(@PathParam("environmentId") String environmentI
...
+    private Response getConfigurationHelper(String environmentId, String feedId, String resourcePath,
+                                               DataEntity.Role dataType) {
+        String tenantId = getTenantId();
+        CanonicalPath resource = composeCanonicalPath(tenantId, environmentId, feedId, resourcePath);
+
+        Resources.Single resourceSingle = inventory.inspect(resource, Resources.Single.class);
+        DataEntity entity;
+        switch (dataType) {
+            case configuration:
+                entity = resourceSingle.configuration().get(null).entity();
```
jkremser: This granularity is imho ok for the REST api for now.. what would be nice to have is an optional query param with JSON path.

metlos: +1 this is a good start and we can expand on it later. The backend support is kinda there so it'd only be a matter of changing the REST API.

hawkular-inventory-json-helper/src/main/java/org/hawkular/inventory/json/mixins/DataEntityMixin.java
```diff
@@ -19,20 +19,26 @@
...
  */
-public class DataEntityMixin {
+public abstract class DataEntityMixin {
 
     @JsonCreator
     public DataEntityMixin(@JsonProperty("path") CanonicalPath path, @JsonProperty("value") StructuredData value,
             @JsonProperty("properties") Map<String, Object> properties) {
     }
 
+    @JsonIgnore abstract String getId();
```
metlos, line 40: So we are only left with `path` in JSON right? I think this makes sense for the new entities we add. I left the `tenantId` and all the other guys in the json ouput because I think someone depended on those attributes being there but I think we should remove them going forward because they are duplicate information.

metlos:
I'll change the `Resources.*.configuration().get(null)` and `connectionConfiguration().get(null)` to `data().get(Role)` as we discussed privately earlier.

This'll enable you to get rid of the ugly switches in your helper methods.

metlos:
Above mentioned change implemented in PR #144.

hawkular-inventory-json-helper/src/main/java/org/hawkular/inventory/json/mixins/DataEntityMixin.java
jkremser, line 40: Without this change the `DataEntity` JSON looks like: 

``` json
{
  "path":"/e;e/r;r/d;connectionConfiguration",
  "value":[1,2],
  "properties":{},
  "role":"connectionConfiguration",
  "id":"connectionConfiguration"
}
```

=> a lot of redundancy
With this change it looks like:

``` json
{
  "path":"/e;e/r;r/d;connectionConfiguration",
  "value":[1,2],
  "properties":{}
}
```

It's more similar to the `DataEntity.{Blueprint|Update}` JSON which are inputs to the modifying ops and contains only the `StructuredData`, i.e. the their JSON looks like:

``` json
{
  "value":[1,2],
}
```

(the `DataEntity.Update` contains also the properties)

jkremser, line 40: btw. I think, we should test the format of the json, currently what we do is checking if 

``` haskell
deserialization . serialization ~ id
```

That's completely ok, but it doesn't say a word about the JSON format

jkremser:
> I'll change the Resources.*.configuration().get(null) and connectionConfiguration().get(null) to data().get(Role) as we discussed privately earlier.
> 
> This'll enable you to get rid of the ugly switches in your helper methods.

I've rebased this PR on #144

hawkular-inventory-api/src/main/java/org/hawkular/inventory/api/model/DataEntity.java
```diff
@@ -125,21 +130,33 @@ public Blueprint build() {
...
+        // this is needed for Jackson deserialization
+        public Update() {
+            this(null, null, null);
+        }
+
         public StructuredData getValue() {
             return value;
         }
 
+        public Role getRole() {
```
metlos: This would imply that one can change the role of the data entity. That is not the case though, so this field should not be needed.

hawkular-inventory-rest-api/src/main/java/org/hawkular/inventory/rest/RestResources.java
```diff
@@ -563,6 +564,183 @@ public Response associateMetrics(@PathParam("environmentId") String environmentI
...
+    @ApiOperation("Updates the configuration of a resource")
+    @ApiResponses({
+                          @ApiResponse(code = 204, message = "OK"),
+                          @ApiResponse(code = 404, message = "Tenant, environment or resource doesn't exist",
+                                       response = ApiError.class),
+                          @ApiResponse(code = 500, message = "Server error", response = ApiError.class)
+                  })
+    public Response updateConfiguration(@PathParam("environmentId") String environmentId,
+                                        @PathParam("resourcePath") String resourcePath,
+                                        @ApiParam(required = true) DataEntity.Update configuration) {
```
metlos, line 687: Aren't you missing `@QueryParam` here?

hawkular-inventory-rest-api/src/main/java/org/hawkular/inventory/rest/RestResources.java
```diff
@@ -563,6 +564,183 @@ public Response associateMetrics(@PathParam("environmentId") String environmentI
...
+    @ApiResponses({
+                          @ApiResponse(code = 204, message = "OK"),
+                          @ApiResponse(code = 404, message = "Tenant, environment, resource or feed doesn't exist",
+                                       response = ApiError.class),
+                          @ApiResponse(code = 500, message = "Server error", response = ApiError.class)
+                  })
+    public Response updateConfiguration(@PathParam("environmentId") String environmentId,
+                                        @PathParam("feedId") String feedId,
+                                        @PathParam("resourcePath") String resourcePath,
+                                        @ApiParam(required = true) DataEntity.Update configuration) {
```
metlos, line 670: Aren't you missing @QueryParam here?

hawkular-inventory-rest-api/src/main/java/org/hawkular/inventory/rest/RestResources.java
```diff
@@ -563,6 +564,183 @@ public Response associateMetrics(@PathParam("environmentId") String environmentI
...
+    @ApiOperation("Creates the configuration for pre-existing resource")
+    @ApiResponses({
+                          @ApiResponse(code = 204, message = "OK"),
+                          @ApiResponse(code = 404, message = "Tenant, environment or resource doesn't exist",
+                                       response = ApiError.class),
+                          @ApiResponse(code = 500, message = "Server error", response = ApiError.class)
+                  })
+    public Response createConfiguration(@PathParam("environmentId") String environmentId,
+                                        @PathParam("resourcePath") String resourcePath,
+                                        @ApiParam(required = true) DataEntity.Blueprint configuration) {
```
metlos: Aren't you missing `@QueryParam` here?

hawkular-inventory-rest-api/src/main/java/org/hawkular/inventory/rest/RestResources.java
```diff
@@ -563,6 +564,183 @@ public Response associateMetrics(@PathParam("environmentId") String environmentI
...
+    @ApiResponses({
+                          @ApiResponse(code = 204, message = "OK"),
+                          @ApiResponse(code = 404, message = "Tenant, environment, resource or feed doesn't exist",
+                                       response = ApiError.class),
+                          @ApiResponse(code = 500, message = "Server error", response = ApiError.class)
+                  })
+    public Response createConfiguration(@PathParam("environmentId") String environmentId,
+                                        @PathParam("feedId") String feedId,
+                                        @PathParam("resourcePath") String resourcePath,
+                                        @ApiParam(required = true) DataEntity.Blueprint configuration) {
```
metlos: Aren't you missing `@QueryParam` here?

hawkular-inventory-rest-api/src/main/java/org/hawkular/inventory/rest/RestResources.java
jkremser: nope, the role is in the blueprint

metlos:
Would you please add tests to `InventoryITest.groovy` in `hawkular/hawkular` to the `dev/inventory-0.3.0Final` branch?

hawkular-inventory-rest-api/src/main/java/org/hawkular/inventory/rest/RestResources.java
jkremser, line 670: probably yes, I put the role into the `DataEntity.Update` but it doesn't make sense, using the `@queryparam` and `DataEntity.Update - role` is better

jkremser:
> Would you please add tests to InventoryITest.groovy in hawkular/hawkular to the dev/inventory-0.3.0Final branch?

Yes, I have the tests locally

hawkular-inventory-rest-api/src/main/java/org/hawkular/inventory/rest/RestResources.java
jkremser: nope, the role is in the blueprint

hawkular-inventory-api/src/main/java/org/hawkular/inventory/api/model/DataEntity.java
```diff
@@ -86,6 +86,11 @@ public Blueprint(DataEntity.Role role, StructuredData value, Map<String, Object>
             this.value = value;
         }
 
+        // this is needed for Jackson deserialization
+        public Blueprint() {
```
metlos: Jackson is fine with private constructors. IMHO this and the Update one should be private or at least protected so that users cannot instantiate this essentially invalid objects.

PR #143, opened 2015-08-12
[HWKINVENT-100] Fix support for resource hierarchy in REST

metlos:
and change the path serialization to use percent encoding so that supplying
the paths to the REST API becomes easier.

Also added missing REST methods for disassociating child resources.

### DIFF
--- a/hawkular-inventory-api/src/main/java/org/hawkular/inventory/api/model/PathSegmentCodec.java
+++ b/hawkular-inventory-api/src/main/java/org/hawkular/inventory/api/model/PathSegmentCodec.java
@@ -1,0 +1,275 @@
+/*
+ * Copyright 2015 Red Hat, Inc. and/or its affiliates
+ * and other contributors as indicated by the @author tags.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.hawkular.inventory.api.model;
+
+import java.nio.ByteBuffer;
+import java.nio.CharBuffer;
+import java.nio.charset.CharacterCodingException;
+import java.nio.charset.Charset;
+import java.nio.charset.CharsetDecoder;
+import java.nio.charset.CharsetEncoder;
+import java.nio.charset.CoderResult;
+
+/**
+ * Used for serializing the paths. Paths segments are URI-encoded.
+ *
+ * <p>This does the URI encoding and decoding roughly twice as fast as {@link java.net.URLEncoder} and
+ * {@link java.net.URLDecoder}.
+ *
+ * @author Lukas Krejci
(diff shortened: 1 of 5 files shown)


jkremser:
So on the agent side this http://git.io/v3EbA is no longer valid. `URLEncoder.encode()` should do the trick, right?

hawkular-inventory-api/src/main/java/org/hawkular/inventory/api/model/PathSegmentCodec.java
```diff
@@ -0,0 +1,275 @@
...
+      /*
+       * <a href="http://ietf.org/rfc/rfc3986.txt">RFC 3986</a>
+       *
+       * unreserved  = ALPHA / DIGIT / "-" / "." / "_" / "~"
+       * sub-delims  = "!" / "$" / "&" / "'" / "(" / ")"
+                     / "*" / "+" / "," / ";" / "="
+       * pchar = unreserved / pct-encoded / sub-delims / ":" / "@"
+       *
+       */
+        return ('a' <= c && c <= 'z') || ('A' <= c && c <= 'Z') || ('0' <= c && c <= '9') ||
```
jkremser, line 272: Could you please change that to `Character.isLetterOrDigit(c)` this is not very readable

check
http://goo.gl/X0HNjG

hawkular-inventory-rest-api/src/main/java/org/hawkular/inventory/rest/RestResources.java
```diff
@@ -28,6 +28,7 @@
 import javax.ws.rs.Consumes;
 import javax.ws.rs.DELETE;
 import javax.ws.rs.DefaultValue;
+import javax.ws.rs.Encoded;
```
jkremser, line 31: Nice! This is what I was looking for

hawkular-inventory-api/src/main/java/org/hawkular/inventory/api/model/PathSegmentCodec.java
metlos, line 272: Nope, that would be incorrect. `Character.isLetterOrDigit(c)` works across the whole unicode spectrum, while this method has to strictly escape everything out of US-ASCII.

Check http://goo.gl/WI3onT ;)

hawkular-inventory-rest-api/src/main/java/org/hawkular/inventory/rest/RestResources.java
metlos, line 31: I was so happy when I found it :)

metlos:
> So on the agent side this http://git.io/v3EbA is no longer valid. URLEncoder.encode() should do the trick, right?

Yes, URL encoding should be everything one needs.

hawkular-inventory-api/src/main/java/org/hawkular/inventory/api/model/PathSegmentCodec.java
jkremser, line 272: well your example is not compilable with the online editor, I guess it hasn't support for 'ř' :), if you want to use ř, you need to use it as \u0159

http://goo.gl/5aXcw6

This should work locally, but doesn't compile on the online tool:

``` java
System.out.println("Přílišžluťoučkýkůňúpělďábelskéódy".chars().allMatch(c -> Character.isAlphabetic(c)));
```

metlos, line 272: Yes, but that's not what I want in my method, because I need to escape any of those characters. So I opted for the "primitive" way of just checking for ranges of chars.

PR #144, opened 2015-08-12
[HWKINVENT-62] Simplify the API.

metlos:
Resources.*.configuration and connectionConfiguration methods replaced
by data() method. The different kinds of data are then determined by the
get(DataEntity.Role) method on the data entity access interfaces.

This change enabled us to get rid of the special relationships for data
roles.

Also renamed Datas -> Data.


jkremser:
Looks good. Data is more abstract and can serve more general purpose in the future. My only comment is that now it could be tricky/confusing for the API consumer what to use: Properties x Data x both x DataWithProperties x Properties + DataWithProperties

metlos:
properties are not queryable and cannot be queried "into" like data can. I think we are now ready to restrict properties to just string values or maybe even re-purpose them as tags or something (although tags would be better modelled differently in the graph db, imho).

PR #145, opened 2015-08-12
these come straight from WildFly MeasurementUnit. This change makes m…

jmazzitelli:
This change makes mapping between them and Hawkular's easy:

MeasurementUnit.valueOf(inventoryMetricUnit.name)

all tests pass with this commit on my local box

This replaces PR #71


jkremser:
lgtm

PR #148, opened 2015-08-13
[HWKINVENT-67] Configuration definitions.

metlos:
The definitions are modelled using JSON schema.
- Split up DataEntity.Role so that its use in different scenarios can be
  type-checked.
- Added ResourceTypes.*.data() to access the data associated with the
  resource type which can be either config or connection config schema.
- Added validation logic for resource data against schemas on the resource
  types
- Added validation logic of the resource type data against json schema
  meta-schema (i.e. if they are well-formed json-schema documents).
- Properly call cleanup and preUpdate checks both in ReadWrite and Single
  impls.
- Added ability to set a starting element for a traversal query to the
  InventoryBackend interface.

### DIFF
--- a/hawkular-inventory-api/pom.xml
+++ b/hawkular-inventory-api/pom.xml
@@ -44,6 +44,12 @@
       <version>1.0.8</version>
     </dependency>
 
+    <dependency>
+      <groupId>com.github.fge</groupId>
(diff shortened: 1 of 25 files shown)


metlos:
REST is missing support for this... One more commit needed ;)

PR #153, opened 2015-08-26
Map ValidationExceptions to HTTP 400 Bad request

ppalaga:


ppalaga:
@metlos with this PR, I am able to fix the integration tests in HK main

PR #154, opened 2015-08-27
[HWKINVENT-130] Add some wait time between transaction retries

metlos:
Add some wait time between transaction retries in an attempt to reduce lock contention on the titan indexes during insertion bursts.

### DIFF
--- a/hawkular-inventory-api/src/main/java/org/hawkular/inventory/base/Util.java
+++ b/hawkular-inventory-api/src/main/java/org/hawkular/inventory/base/Util.java
@@ -26,6 +26,7 @@ import static org.hawkular.inventory.api.Relationships.WellKnown.contains;
 import static org.hawkular.inventory.api.Relationships.WellKnown.defines;
 
 import java.util.HashSet;
+import java.util.Random;
 import java.util.Set;
 import java.util.function.BiConsumer;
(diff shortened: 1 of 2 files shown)


hawkular-inventory-api/src/main/java/org/hawkular/inventory/base/Util.java
```diff
@@ -86,12 +92,34 @@ private Util() {
...
 
                 lastException = e;
+
+                if (failures < maxFailures) {
+                    try {
+                        Thread.sleep(waitTime);
+                    } catch (InterruptedException ie) {
+                        Log.LOGGER.wInterruptedWhileWaitingForTransactionRetry();
+                        //reset the interruption flag
+                        Thread.currentThread().interrupt();
```
jkremser, line 109: This is here because the `InterruptedException` clears the flag, right?

metlos, line 109: yes

metlos, line 109: Throwing the transaction failure afterwards is maybe a little bit problematic, but I think necessary, because that's the only possible way of telling the caller that we weren't able to commit successfully.

PR #157, opened 2015-09-01
HWKINVENT-132 - Rest endpoint for retrieving arbitrary entity by path is broken

jkremser:
- extracting the entity path from @UriInfo including the escaped characters (the false param)


metlos:
Yay!

PR #158, opened 2015-09-01
[HWKINVENT-128] Operation types

metlos:
This introduces the operation types as entities contained in resource
types.

Unlike metric types that are merely associated with the resource types but
otherwise are independent, the operation types are truly contained within
resource types and cannot live without it.

The thinking behind that is that unlike metrics, operations really only
make sense on resources. It is not particularly useful to think about
resourceless operations executed directly by the feed. Or rather the cost
of having such standalone operations and having to address them uniquely
within the feed or globally in a tenant outweighs the benefit of having
them.

As such the operation types are unique to the resource types.

~~WARNING: this commit does NOT address operation types in the REST API.~~


metlos:
Rebased on top of the latest master.

hawkular-inventory-api/src/test/java/org/hawkular/inventory/api/test/AbstractBaseInventoryPersistenceCheck.java
```diff
@@ -1629,52 +1676,52 @@ public void testBackendFind() throws Exception {
         CanonicalPath envPath = tenantPath.extend(Environment.class, "production").get();
         entity = backend.find(envPath);
         Environment env = backend.convert(entity, Environment.class);
-        Assert.assertEquals("com.acme.tenant", env.getTenantId());
+        Assert.assertEquals("com.acme.tenant", env.getPath().ids().getTenantId());
```
jkremser, line 1679: are these

``` java
getXId() -> getPath().ids().getXId()
```

 the only breaking changes, or is the `getXId()` method still there?

metlos, line 1679: Those methods are gone. I think this will also break the REST API because those fields will no longer be present in the serialized form.

metlos, line 1679: I actually don't know if anything else broke API. Let's use revapi to find out :)

jkremser:
So both resource types under feed and feedless ones can now contain the operation type node that can contain 2 kinds of JSON schema, 1 for params and 1 for the return type, right?

metlos:
aye

PR #161, opened 2015-09-07
Adding a way to completelly disable the Auth{N|Z}

jkremser:
To completely disable the auth completely do following:
- Use the `@AllPermissive` annotation in `RestBase` class (it's commented there)
- Set the DUMMY flag to true in `SecurityIntegration` class (it's commented there)
- Comment out the secure-deployment element for hawkular-inventory-dist.war in `dist/src/main/resources/wildfly/patches/standalone.xsl` in `hawkular/hawkular` repo (http://git.io/vZkQs)
- Comment out the keycloak xml elements in `jboss-web.xml` `web.xml` (it's commented there)

**NOTE:** it can be disabled also only partially, for instance if the
`@AllPermissive` annotation is used only above the `tenantIdProducer` field
the auth will be working but the tenant will be always the same.

**NOTE2:** This shouldn't be understood that we want to disable the auth altogether. Rather, it's useful to measure the impact of security aspects to the overall performance.


hawkular-inventory-rest-api/src/main/webapp/WEB-INF/jboss-web.xml
```diff
@@ -19,5 +19,8 @@
 -->
 <jboss-web>
   <context-root>/hawkular/inventory</context-root>
-  <security-domain>keycloak</security-domain>
+
+  <!-- comment the following line out to disable keycloak auth -->
+  <!--<security-domain>keycloak</security-domain>-->
```
metlos: Should this really be left commented in the default state? I think we should by default function authz'ed.

metlos:
Are the instructions on how to switch to unsecured inventory listed somewhere in the source code? Would be good for future reference, even if it gets outdated ;)

hawkular-inventory-rest-api/src/main/webapp/WEB-INF/jboss-web.xml
jkremser: right, it shouldn't

jkremser:
> Are the instructions on how to switch to unsecured inventory listed somewhere in the source code? Would be good for future reference, even if it gets outdated ;)

Yes, but no at 1 place. It's in the commit msg and the PR description. I guess, I could put it also to `SecurityIntegration` class jdoc.

PR #166, opened 2015-09-16
[HWKINVENT-122] Bulk Resource creation

metlos:
This is only a part of HWKINVENT-122, we're missing the "relate" part but that's already being worked on. I wanted to do a proof of concept on one type of entity to see how it performs.

### DIFF
--- a/hawkular-inventory-api/src/main/java/org/hawkular/inventory/api/model/Entity.java
+++ b/hawkular-inventory-api/src/main/java/org/hawkular/inventory/api/model/Entity.java
@@ -16,9 +16,11 @@
  */
 package org.hawkular.inventory.api.model;
 
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.HashSet;
 import java.util.Map;
-
-import javax.xml.bind.annotation.XmlAttribute;
+import java.util.Set;
 
 /**
  * Base class for all Hawkular entities.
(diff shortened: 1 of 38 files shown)


metlos:
Rebased on the latest master

hawkular-inventory-api/src/main/java/org/hawkular/inventory/api/model/Entity.java
```diff
@@ -72,27 +74,89 @@ public final String toString() {
      * the subclasses are free to use a blueprint type not inheriting from this one.
      */
     public abstract static class Blueprint extends AbstractElement.Blueprint {
-        @XmlAttribute
         private final String id;
+        private final Map<String, Set<CanonicalPath>> outgoing;
```
jkremser, line 78: So if I understand it correctly, one can specify the relationships in the entity blueprint itself and it will be automatically wired when creating this entity. I.e. 1 call instead of N (?)

metlos, line 78: Yes

metlos, line 78: ~~But because I forgot that I actually added this in, I didn't add tests for it and _gasp_ I didn't even implement it ;)~~

UPDATE: implemented

metlos:
Adding more tests, discovering issues here and there...

metlos:
rebased on the latest master again.

hawkular-inventory-rest-api/src/main/java/org/hawkular/inventory/rest/RestBulk.java
```diff
@@ -0,0 +1,510 @@
...
+                    } else {
+                        bulkCreateEntity(statuses, idExtractor, parentPath, single, elementType, blueprints);
+                    }
+                }
+            }
+            transaction.commit();
+            return statuses;
+        } catch (Throwable t) {
+            // TODO this potentially leaves behind the security resources of the entities that have been created
+            // before the transaction failure
```
jkremser, line 375: it looks like `bulkCreateEntity()` swallows all the exceptions, perhaps `bulkCreateRelationships()` could have the same Exception -> http 500 thing.

One thinkable way of reaching this place is the situation when the path is wrong (line 355). Then the inner for-loop doesn't make sense.. Perhaps, setting all the status codes new entities on this non-existent path to 404 could work.

Actually, I think this method should be as robust as possible and the rollback is probably an overkill. Btw. is it possible to create an entity and in the same bulk operation also another entity under it (parent-child resources for instance)?

PR #168, opened 2015-09-23
Introduce API checking using Revapi

metlos:
\+ a couple of visibility changes for problems discovered by Revapi.

Note that the versions of the Revapi components in this PR are not yet available in Maven central, so we'll have to wait a bit before this passes CI. It's passing locally.


metlos:
Hmm, so for a reason not completely clear to me, the plugin determines that the last released version of hawkular-inventory-json-helper and hawkular-inventory-cdi is 0.2.0.Alpha1 which it is not. Not sure what might be going on there.

PR #172, opened 2015-10-01
[HWKINVENT-95] Disallow manual deletion and creation of "defines"

metlos:
relationships so that it is not possible to manually break the contract of
resources and metrics that are always supposed to be defined by their
respective type.


jkremser:
looks safe

PR #173, opened 2015-10-01
Log the stacktrace on debug when inventory init fails.

metlos:
The `debugf()` call made previously didn't do what we thought it'd do.


jkremser:
:+1:

pilhuhn:
See https://issues.jboss.org/browse/HWKINVENT-147

PR #176, opened 2015-10-05
Make EntityType enum key names derivable from Entity class names

ppalaga:


ppalaga:
@metlos would you please review and eventually merge?

PR #178, opened 2015-10-06
Hwkinvent-149

pavolloffay:
Move inventory integration tests to inventory repository.

### DIFF
--- a/hawkular-inventory-itest-parent/hawkular-inventory-itest-dist/pom.xml
+++ b/hawkular-inventory-itest-parent/hawkular-inventory-itest-dist/pom.xml
@@ -1,0 +1,531 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+
+    Copyright 2015 Red Hat, Inc. and/or its affiliates
+    and other contributors as indicated by the @author tags.
+
+    Licensed under the Apache License, Version 2.0 (the "License");
+    you may not use this file except in compliance with the License.
+    You may obtain a copy of the License at
+
+       http://www.apache.org/licenses/LICENSE-2.0
+
+    Unless required by applicable law or agreed to in writing, software
+    distributed under the License is distributed on an "AS IS" BASIS,
+    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+    See the License for the specific language governing permissions and
+    limitations under the License.
+
+-->
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+  <modelVersion>4.0.0</modelVersion>
+
+  <parent>
+    <groupId>org.hawkular.inventory</groupId>
+    <artifactId>hawkular-inventory-itest-parent</artifactId>
+    <version>0.6.0-SNAPSHOT</version>
+  </parent>
+
+  <artifactId>hawkular-inventory-itest-dist</artifactId>
+
+  <properties>
+    <kettle.build.type>production</kettle.build.type>
+    <assembly.dir>${project.build.directory}/assembly</assembly.dir>
+    <wildfly.dist.zip.root.dir>wildfly-${version.org.wildfly}</wildfly.dist.zip.root.dir>
+    <hawkular.dist.zip.root.dir>${project.artifactId}-${project.version}</hawkular.dist.zip.root.dir>
+  </properties>
+
+  <build>
+    <plugins>
+      <!-- skip tests -->
+      <plugin>
+        <groupId>org.apache.maven.plugins</groupId>
+        <artifactId>maven-surefire-plugin</artifactId>
+        <configuration>
+          <skipTests>true</skipTests>
+        </configuration>
+      </plugin>
+    </plugins>
+  </build>
+
+  <dependencies>
+    <dependency>
+      <groupId>org.hawkular.inventory</groupId>
+      <artifactId>hawkular-inventory-itest</artifactId>
+      <version>${project.version}</version>
+    </dependency>
+
+    <!-- test dist dependencies -->
+    <dependency>
+      <groupId>org.hawkular.bus</groupId>
+      <artifactId>hawkular-bus-broker-wf-extension</artifactId>
+      <version>${test-dist-version.org.hawkular.bus}</version>
+      <classifier>module</classifier>
+      <type>zip</type>
+    </dependency>
+    <dependency>
+      <groupId>org.hawkular.bus</groupId>
+      <artifactId>hawkular-bus-ra-wf-module</artifactId>
+      <version>${test-dist-version.org.hawkular.bus}</version>
+      <classifier>module</classifier>
+      <type>zip</type>
+    </dependency>
+    <dependency>
+      <groupId>org.hawkular.nest</groupId>
+      <artifactId>hawkular-nest-wf-extension</artifactId>
+      <version>${test-dist-version.org.hawkular.bus}</version>
+      <classifier>module</classifier>
+      <type>zip</type>
+    </dependency>
+    <dependency>
+      <groupId>org.hawkular.inventory</groupId>
+      <artifactId>hawkular-inventory-dist</artifactId>
+      <version>${project.version}</version>
+      <type>war</type>
+    </dependency>
+    <dependency>
+      <groupId>org.hawkular.commons</groupId>
+      <artifactId>hawkular-commons-embedded-cassandra-ear</artifactId>
+      <version>${test-dist-version.org.hawkular.commons}</version>
+      <type>ear</type>
+    </dependency>
+    <dependency>
+      <groupId>org.hawkular.accounts</groupId>
+      <artifactId>hawkular-accounts</artifactId>
+      <version>${test-dist-version.org.hawkular.accounts}</version>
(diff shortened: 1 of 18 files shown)


metlos:
Could you also copy over the `Scenario1.groovy` (and call it something different) from hawkular/hawkular? It contains some tests that are not mirrored in the `InventoryITest` and I'd like to have the complete set of itests moved to the component.

pavolloffay:
Sure, I will copy it.

pavolloffay:
@ppalaga could you please review

pavolloffay:
Now it is similar to the solution in agent. 

Scenario1 was removed it depends on Metrics. I will send PR to hawkular which will reuse tests from inventory. It will need a release of tests.

pavolloffay:
@metlos @Jiri-Kremser to run itests please run mvn clean verify -Ditest, itest profile is divided into 2 smaller profiles (compose dist, run tests)

hawkular-inventory-itest-parent/hawkular-inventory-itest-dist/pom.xml
```diff
@@ -0,0 +1,592 @@
...
+      <groupId>org.hawkular.inventory</groupId>
+      <artifactId>hawkular-inventory-itest</artifactId>
+      <version>${project.version}</version>
+    </dependency>
+
+    <!-- test dist dependencies -->
+    <dependency>
+      <groupId>org.hawkular.bus</groupId>
+      <artifactId>hawkular-bus-broker-wf-extension</artifactId>
+      <version>${test-dist-version.org.hawkular.bus}</version>
```
ppalaga, line 64: Why not simply `test-dist-version`?

ppalaga, line 64: Esp. in case of bus, you want to test against the same version as the rest of inventory depends on, do you not?

hawkular-inventory-itest-parent/hawkular-inventory-itest-dist/pom.xml
```diff
@@ -0,0 +1,592 @@
...
+      <groupId>org.hawkular.nest</groupId>
+      <artifactId>hawkular-nest-wf-extension</artifactId>
+      <version>${test-dist-version.org.hawkular.bus}</version>
+      <classifier>module</classifier>
+      <type>zip</type>
+    </dependency>
+    <dependency>
+      <groupId>org.hawkular.inventory</groupId>
+      <artifactId>hawkular-inventory-dist</artifactId>
+      <version>${test-dist-version.org.hawkular.inventory}</version>
```
ppalaga: this should be simply `${project.version}` because you'll never ever will want to test against anything else than the current inventory version.

hawkular-inventory-itest-parent/hawkular-inventory-itest-dist/pom.xml
```diff
@@ -0,0 +1,592 @@
...
+    <dependency>
+      <groupId>org.hawkular.commons</groupId>
+      <artifactId>hawkular-commons-embedded-cassandra-ear</artifactId>
+      <version>${test-dist-version.org.hawkular.commons}</version>
+      <type>ear</type>
+    </dependency>
+    <dependency>
+      <groupId>org.hawkular.accounts</groupId>
+      <artifactId>hawkular-accounts</artifactId>
+      <version>${test-dist-version.org.hawkular.accounts}</version>
```
ppalaga, line 97: same as bus. I.e. here^ as well as in rest-api https://github.com/hawkular/hawkular-inventory/blob/master/hawkular-inventory-rest-api/pom.xml#L106 , `${version.org.hawkular.accounts} should be used that in turn should be defined here around here in top leve inv parent: https://github.com/hawkular/hawkular-inventory/blob/master/pom.xml#L56

I see no reason for using `test-dist-` prefix even for `commons` you use it and you depend on it. No matter that it is in the test. There is only one convention for version prop names - they should start with `version.org.hawkuar...`

hawkular-inventory-itest-parent/hawkular-inventory-itest-dist/pom.xml
```diff
@@ -0,0 +1,592 @@
...
+                <goals>
+                  <goal>single</goal>
+                </goals>
+              </execution>
+            </executions>
+          </plugin>
+
+          <plugin>
+            <groupId>com.spotify</groupId>
+            <artifactId>docker-maven-plugin</artifactId>
```
ppalaga: I'd say that `docker-maven-plugin` should not be here, as it should not be in agent-itest-dist

hawkular-inventory-itest-parent/hawkular-inventory-itest-dist/pom.xml
```diff
@@ -0,0 +1,592 @@
...
+      <!--<artifactId>hawkular-command-gateway-war</artifactId>-->
+      <!--<version>${version.org.hawkular.cmdgw}</version>-->
+      <!--<type>war</type>-->
+    <!--</dependency>-->
+    <!--<dependency>-->
+      <!--<groupId>org.hawkular.metrics</groupId>-->
+      <!--<artifactId>hawkular-metrics-api-jaxrs</artifactId>-->
+      <!--<version>${version.org.hawkular.metrics}</version>-->
+      <!--<type>war</type>-->
+    <!--</dependency>-->
```
ppalaga: I do not think that there is a chance that we will need these commented lines in the future

hawkular-inventory-itest-parent/hawkular-inventory-itest-dist/pom.xml
```diff
@@ -0,0 +1,592 @@
...
+      <artifactId>hawkular-accounts</artifactId>
+      <version>${test-dist-version.org.hawkular.accounts}</version>
+      <type>war</type>
+    </dependency>
+    <!--<dependency>-->
+      <!--<groupId>org.hawkular.accounts</groupId>-->
+      <!--<artifactId>hawkular-accounts-events-backend</artifactId>-->
+      <!--<version>${version.org.hawkular.accounts}</version>-->
+      <!--<type>war</type>-->
+    <!--</dependency>-->
```
ppalaga: Remove `hawkular-accounts-events-backend` altogether if the distro works without it.

hawkular-inventory-itest-parent/hawkular-inventory-itest-dist/pom.xml
```diff
@@ -0,0 +1,592 @@
...
+                </configuration>
+              </execution>
+            </executions>
+          </plugin>
+        </plugins>
+      </build>
+    </profile>
+
+    <profile>
+      <id>run-test-dist-and-tests</id>
```
ppalaga: Why do we need to keep `run-test-dist-and-tests` profile separate from `compose-test-dist` ? Esp. if they are both triggered by the same property, hence the always run together?

pom.xml
```diff
@@ -54,7 +55,14 @@
 
   <properties>
     <version.org.hawkular.bus>0.7.0.Final</version.org.hawkular.bus>
+    <version.org.hawkular.accounts>1.0.14.Final</version.org.hawkular.accounts>
```
pavolloffay, line 58: @metlos  I moved accounts prop from rest-api here.

pom.xml
```diff
@@ -54,7 +55,14 @@
...
     <api-check-activated>true</api-check-activated>
+
+    <!-- integration tests distribution properties -->
+    <test-dist-version.org.hawkular.bus>${version.org.hawkular.bus}</test-dist-version.org.hawkular.bus>
+    <test-dist-version.org.hawkular.accounts>${version.org.hawkular.accounts}</test-dist-version.org.hawkular.accounts>
+    <test-dist-version.org.hawkular.commons>0.2.1.Final</test-dist-version.org.hawkular.commons> <!-- C* -->
+    <test-dist-version.org.keycloak>1.3.1.Final</test-dist-version.org.keycloak>
```
pavolloffay, line 65: Is it ok like this? I think is better to be able to change version of accounts, bus for test distribution.

metlos, line 65: :+1: from me for this ability. Good that the default version is shared between the itest and inventory-bus components.

pom.xml
metlos, line 58: :+1:

hawkular-inventory-itest-parent/hawkular-inventory-itest-dist/pom.xml
metlos, line 97: I think the ability to potentially replace the version of the components used in the itest is good for debugging reasons.

hawkular-inventory-itest-parent/hawkular-inventory-itest-dist/pom.xml
metlos, line 64: By default, yes. But I think it'd be good to be able to replace that when for example testing locally before switching to a new version that hawkular dist started using, without the need to do a full dance with build hawkular dist with the snapshot of inventory.

pom.xml
ppalaga, line 65: @pavolloffay could you please add `"hawkular-inventory" -> "hawkular-commons" [style=dashed, color=gray, label="itest" fontcolor="grey"]` around here https://github.com/hawkular/hawkular.github.io/blob/pages/src/main/manually-generated/assets/img/docs/dev/components-dependencies.dot#L71 once this is merged?

It is the source for this dependency graph: http://www.hawkular.org/docs/dev/development.html#component-dependencies . I hope nothing else is missing in the graph.

pavolloffay, line 65: Done https://github.com/hawkular/hawkular.github.io/pull/107

PR #179, opened 2015-10-06
Fixing the streaming serialiaztion. 

jkremser:
There was a timing issue resulting in calling `Page.getTotalSize()` on non-closed page -> NPE

### DIFF
--- a/hawkular-inventory-api/src/main/java/org/hawkular/inventory/api/paging/SizeAwarePage.java
+++ b/hawkular-inventory-api/src/main/java/org/hawkular/inventory/api/paging/SizeAwarePage.java
@@ -50,7 +50,7 @@ public class SizeAwarePage<T> extends Page<T> {
     @Override
     public long getTotalSize() {
         if (totalSize == HasTotalSize.NOT_DEPLETED && !hasNext()) {
-            totalSize = hasTotalSize.getTotalSize();
+            totalSize = hasTotalSize == null ? HasTotalSize.NOT_DEPLETED : hasTotalSize.getTotalSize();
         }
         return totalSize;
     }
--- a/hawkular-inventory-rest-api/src/main/java/org/hawkular/inventory/rest/ResponseUtil.java
+++ b/hawkular-inventory-rest-api/src/main/java/org/hawkular/inventory/rest/ResponseUtil.java
@@ -27,6 +27,7 @@ import java.util.List;
 import java.util.Spliterator;
 import java.util.stream.StreamSupport;
 
+import javax.ws.rs.core.MediaType;
 import javax.ws.rs.core.Response;
 import javax.ws.rs.core.UriBuilder;
 import javax.ws.rs.core.UriInfo;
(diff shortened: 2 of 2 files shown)


hawkular-inventory-api/src/main/java/org/hawkular/inventory/api/paging/SizeAwarePage.java
```diff
@@ -50,7 +50,7 @@ public SizeAwarePage(Iterator<T> wrapped, PageContext pageContext, HasTotalSize
     @Override
     public long getTotalSize() {
         if (totalSize == HasTotalSize.NOT_DEPLETED && !hasNext()) {
-            totalSize = hasTotalSize.getTotalSize();
+            totalSize = hasTotalSize == null ? HasTotalSize.NOT_DEPLETED : hasTotalSize.getTotalSize();
```
metlos, line 53: Technically, setting `totalSize` to `NOT_DEPLETED` here is not correct, right? Because in this block we know we've depleted the iterator, it's just that `hasTotalSize` has already been nulled out (which I guess can only happen when you somehow call `close()` and `hasNext()` still returns true).

metlos:
So with this, the streaming approach works? Shouldn't we then switch to it by default?

jkremser:
> So with this, the streaming approach works? Shouldn't we then switch to it by default?

Yes it works, however, there is no performance gain, now. The paging must be killing the performance, though, because I was trying:

``` bash
curl -iv -u jdoe:password http://localhost:8080/hawkular/inventory/test/localhost/resources/localhost~Local~~/children -w "\\n CN: %{time_connect}\\n ST: %{time_starttransfer}\\n TT: %{time_total}"
```

vs

``` bash
curl -iv -u jdoe:password http://localhost:8080/hawkular/inventory/test/localhost/resources/localhost~Local~~/recursiveChildren -w "\\n CN: %{time_connect}\\n ST: %{time_starttransfer}\\n TT: %{time_total}"
```

(/children vs /recursiveChildren) and the latter is far more performant. The first call uses paging the second one does not.

EDIT: oh, and the reason I turned it off is because guys on Mac have (imho) problems with performance because of the multi-threading sux on Macbooks (spawning new threads won't help them)

hawkular-inventory-api/src/main/java/org/hawkular/inventory/api/paging/SizeAwarePage.java
jkremser, line 53: > it's just that `hasTotalSize` has already been nulled out (which I guess can only happen when you somehow call `close()` and `hasNext()` still returns true).

Yes, that was the case, the close() was called prematurely. I fixed it so it should work also w/o this check, I just wanted to be defensive here not allowing to throw the NPE. Perhaps checking the hasTotalSize for null and throwing IllegalStateExc. is better.

Now, I see the commit message is wrong, it should say: "issue resulting in calling Page.getTotalSize() on non-depleted but closed page -> NPE"

PR #181, opened 2015-10-13
[HWKINVENT-143] Add "name" as a first class property on entities.

metlos:


jmazzitelli:
when this is in, let me know and I'll incorporate it via https://issues.jboss.org/browse/HWKAGENT-22

jkremser:
@jmazzitelli it is in :D

hawkular-inventory-api/src/main/java/org/hawkular/inventory/api/model/DataEntity.java
```diff
@@ -141,7 +147,7 @@ static Role valueOf(String name) {
         }
 
         public Blueprint(DataRole role, StructuredData value, Map<String, Object> properties) {
-            super(properties);
+            super(role.name(), properties);
```
jmazzitelli, line 150: this is called from the no-arg constructor (see below line#168). role will always be null, thus, NPE.

hawkular-inventory-api/src/main/java/org/hawkular/inventory/api/model/DataEntity.java
```diff
@@ -149,6 +155,14 @@ public Blueprint(DataRole role, StructuredData value, Map<String, Object> proper
...
+                         Map<String, Set<CanonicalPath>> outgoing,
+                         Map<String, Set<CanonicalPath>> incoming) {
+            super(role.name(), properties, outgoing, incoming);
+            this.role = role;
+            this.value = value;
+        }
+
         // this is needed for Jackson deserialization
         private Blueprint() {
             this(null, null, null);
```
jmazzitelli, line 168: causes NPE, see line 150

hawkular-inventory-api/src/main/java/org/hawkular/inventory/api/model/DataEntity.java
jkremser, line 150: The constructor is private (it's there because of jackson). Have you seen the NPE from the agent?

jmazzitelli, line 150: yes, see my JIRA: https://issues.jboss.org/browse/HWKINVENT-154

PR #183, opened 2015-10-15
[HWKINVENT-151] Basic impl of MetadataPacks

metlos:
Basic implementation of metadatapacks as specified in https://issues.jboss.org/browse/HWKINVENT-92 and https://issues.jboss.org/browse/HWKINVENT-151.

Todo:
- [x] more unit tests (in `AbstractBaseInventoryPersistenceCheck`)
- [x] javadocs
- [x] disallow updates that would influence the hash of the metadatapack
- [x] Rest API


metlos:
Merged with master and is mergeable now.

hawkular-inventory-api/src/main/java/org/hawkular/inventory/base/BaseMetricTypes.java
```diff
@@ -81,7 +84,41 @@ protected String getProposedId(MetricType.Blueprint blueprint) {
...
+
+        private static <BE> void preUpdate(TraversalContext<BE, ?> context, BE entity, MetricType.Update update) {
+            MetricType mt = context.backend.convert(entity, MetricType.class);
+            if (mt.getUnit() == update.getUnit()) {
+                //k, this is the only updatable thing that influences metadata packs, so if it is equal, we're ok.
+                return;
+            }
+
+            if (isInMetadataPack(context, entity)) {
+                throw new IllegalArgumentException("Cannot delete a metric type that is a part of metadata pack.");
```
jkremser: s/delete/update/

hawkular-inventory-api/src/main/java/org/hawkular/inventory/base/BaseMetricTypes.java
```diff
@@ -81,7 +84,41 @@ protected String getProposedId(MetricType.Blueprint blueprint) {
...
+        }
+
+        @Override
+        protected void preUpdate(String s, BE entityRepresentation, MetricType.Update update) {
+            preUpdate(context, entityRepresentation, update);
+        }
+
+        private static <BE> void cleanup(TraversalContext<BE, ?> context, BE deletedEntity) {
+            if (isInMetadataPack(context, deletedEntity)) {
+                throw new IllegalArgumentException("Cannot delete a metric type that is a part of metadata pack.");
```
jkremser, line 102: s/delete/update/

jkremser:
General question: what shoud I do if I created a metadata pack from a bunch of stuff that will change in the future. I understand that the metadata pack form some form of contract, but still. Should I delete the pack, change the underlying stuff and create a new one? What about having something like `--force`?

jkremser:
It's huge amount of work :+1:

metlos:
> General question: what shoud I do if I created a metadata pack from a bunch of stuff that will change in the future. I understand that the metadata pack form some form of contract, but still. Should I delete the pack, change the underlying stuff and create a new one? What about having something like --force?

I see metadata packs as a form of sharing the metadata.  To achieve that they are **addressed** by their hash, which has 2 consequences - if you change their constituent parts, the ID of the metadata pack would need to change, too, which defeats the purpose of an ID ;) but more importantly that precise fact enables the users to be certain that types with certain ids and structure are on places they expect. All that just by checking the ID of the metadata pack.

So if you intent to change the types in the pack, you just delete the pack, which unfreezes the types, do the changes, and then create a new pack. Users that expect the old version will see that it is no longer available and users that use the new version will be able to use the pack.

Now the question comes "what to do when we need both old and new to co-exist?" I think that is, as many things, going to be solved by the entity versioning that is going to come in the future. In that scenario, the old pack could remain but point to the old versions of the types (and the types wouldn't have to be frozen at all).

hawkular-inventory-api/src/main/java/org/hawkular/inventory/base/BaseMetricTypes.java
metlos, line 102: `cleanup()` is called during delete so I think this one's actually ok.

metlos:
I think this is ready to be merged feature-wise. We'll see how this goes and will make amendments based on feedback. @Jiri-Kremser would you please review the new commits?

jkremser:
lgtm
last thing: using the hash as the ID of metadata pack can actually lead to the state, where 2 different metadata packs have the same hash (hash collision). However, the probability is really, really small. sha1 produces 160 bits long digest, so it's 2^160  => cajk :]

PR #184, opened 2015-10-19
[HWKINVENT-154] Fix JSON serialization issues with entity blueprints.

metlos:


hawkular-inventory-api/src/main/java/org/hawkular/inventory/api/model/DataEntity.java
```diff
@@ -147,29 +148,30 @@ static Role valueOf(String name) {
...
         }
 
         public Blueprint(DataRole role, StructuredData value, Map<String, Object> properties,
                          Map<String, Set<CanonicalPath>> outgoing,
                          Map<String, Set<CanonicalPath>> incoming) {
             super(role.name(), properties, outgoing, incoming);
+            if (value == null) {
+                value = StructuredData.get().undefined();
+            }
             this.role = role;
```
jkremser, line 161: can the `role` field be null?

metlos, line 161: no. It should never be null, because it dubs for an ID. Basically, data entities cannot have an arbitrary ID but 1 of an enumerated list defined by the entity type that are being added under.

PR #189, opened 2015-11-04
Hwkinvent 76

jkremser:
This PR depends on #174, so the web UI fails to display the real changes, but it can be seen here cd665430d8f5e


metlos:
PR #174 is now passing. You can rebase on top of it to make this pass, too, before we can merge.

PR #190, opened 2015-11-09
[HWKINVENT-65] Feeds moved from environments to tenants: reflecting t…

jkremser:
…he change for the resource configuration and resource metrics


metlos:
Thanks Jirka for spotting this.

PR #192, opened 2015-11-16
HWKINVENT-160 | collectionInterval to Metric and MetricTypes

pavolloffay:
- Added interval field to MetricType and Metric.  
- Interval represents period in seconds at which metrics are collected. 
- In MetricType this attribute is mandatory. 
- On Metric update (or PUT) when interval is null it is deleted.

### DIFF
--- a/hawkular-inventory-api/src/main/java/org/hawkular/inventory/api/model/Metric.java
+++ b/hawkular-inventory-api/src/main/java/org/hawkular/inventory/api/model/Metric.java
@@ -35,6 +35,7 @@ import javax.xml.bind.annotation.XmlRootElement;
 public final class Metric extends Entity<Metric.Blueprint, Metric.Update> {
 
     private final MetricType type;
+    private final Long collectionInterval;
 
     /**
      * JAXB support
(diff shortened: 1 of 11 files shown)


hawkular-inventory-json-helper/src/test/java/org/hawkular/inventory/json/SerializationTest.java
```diff
@@ -184,7 +184,7 @@ public void testDetypedMetricType() throws Exception {
         MetricType mt = new MetricType(CanonicalPath.fromString("/t;t/mt;c"), MetricUnit.BYTES, GAUGE,
                 new HashMap<String, Object>() {{
                     put("a", "b");
-                }});
+                }}, 0L);
 
         testDetyped(mt, "{\"path\":\"/t;t/mt;c\",\"properties\":{\"a\":\"b\"},\"unit\":\"BYTES\"}");
```
metlos: Shouldn't the interval manifest itself in the json, too?

hawkular-inventory-api/src/main/java/org/hawkular/inventory/api/model/Metric.java
```diff
@@ -35,6 +35,7 @@
 public final class Metric extends Entity<Metric.Blueprint, Metric.Update> {
 
     private final MetricType type;
+    private final Long interval;
```
metlos: I'd maybe be more explicit here and actually call this `collectionInterval`. "Interval of a metric" sounds a bit ambiguous to me. I'd intuitively understand it as the min-max interval of values, not a collection interval.

pavolloffay: Good idea, collectionInterval is more explicit.

pavolloffay:
Renamed.
I have squashed it into 1 commit since I have changed the names in both classes.

hawkular-inventory-api/src/main/java/org/hawkular/inventory/api/model/MetricType.java
```diff
@@ -41,6 +41,9 @@
     @XmlAttribute
     private final MetricDataType type;
 
+    @XmlAttribute
+    private final Long collectionInterval;
```
pilhuhn, line 45: What is the unit?

pilhuhn:
Does / will setting this have any effect on H-metrics or the agent?

pavolloffay:
It is in seconds, I think sync with agent is in Lukas's pipeline.

PR #198, opened 2015-11-26
[HWKIVENT-159] Enable metrics ownership by resources

metlos:
In addition to the fact that metrics can now be also contained within resources themselves (not just associated), this PR brings in a new more powerful and future-proof way of expressing traversals across multiple possibly recursive paths.

Apart from tenant, feed and environment, all other entity types can now be stored on multiple places in inventory - metric and resource types under tenant or feed, resources and metrics under environment or feed and resources can even be hierarchical.

Expressing the individual possible ways using separate methods would lead to a slight combinatorial explosion of possibilities.

As a (partial) solution to this problem the new `<<entityType>>Under(Parents)` methods where introduced using which one can say that they want e.g. resource types both directly under a tenant and under any feed under that tenant or to get all metrics stored in an environment, any associated feed and all resources either under the environment or any of the feeds, recursively.

Examples:
- all the metrics contained in an env, its feeds and any of the resources, recursively

``` java
inventory.inspect(env).metricsUnder(Parents.any());
```
- all metrics in an env and its feeds, but not resources:

``` java
inventory.inspect(env).metricsUnder(Environments.MetricParents.ENVIRONMENT, Environments.MetricParents.FEED);
```
- metrics owned by some resource in an environment or any of its feeds

``` java
inventory.inspect(env).metricsUnder(Environments.MetricParents.RESOURCE);
```


hawkular-inventory-rest-api/src/main/java/org/hawkular/inventory/rest/RestResourcesMetrics.java
```diff
@@ -64,17 +67,26 @@
...
-            @Encoded @PathParam("resourcePath") String resourcePath,
-            @ApiParam("A list of paths to metrics to be associated with the resource. They can either be canonical or" +
-                    " relative to the resource.") Collection<String> metricPaths) {
+    public Response associateOrCreateMetrics(@PathParam("environmentId") String environmentId,
+                                             @Encoded @PathParam("resourcePath") String resourcePath,
+                                             @ApiParam("This is either a metric blueprint or a list of paths to " +
+                                                     "metrics to be associated with the resource. They can either be " +
+                                                     "canonical or relative to the resource.")
+                                                 Object metricPathsOrBlueprint,
+                                             @Context UriInfo uriInfo) {
```
jkremser, line 89: does this `Collection<String> metricPaths` -> `Object metricPathsOrBlueprint` change break the rest API?

hawkular-inventory-itest-parent/hawkular-inventory-itest/src/main/groovy/org/hawkular/inventory/rest/test/InventoryITest.groovy
```diff
@@ -734,7 +732,7 @@ class InventoryITest extends AbstractTestBase {
     }
 
     @Test
-    void testResourceTypesOwnMetricTypes() {
+    void testResourceTypesIncorporatesMetricTypes() {
         assertRelationshipExists("resourceTypes/$pingableHostRTypeId/relationships",
                 "/t;$tenantId/rt;$pingableHostRTypeId".toString(),
                 incorporates.name(),
```
jkremser, line 738: note: this will conflict with #201

hawkular-inventory-rest-api/src/main/java/org/hawkular/inventory/rest/RestResourcesMetrics.java
metlos, line 89: No. If you look at the new `createOrAssociateMetric()`, you'll see that it basically is a switch that 1) handles a list of metric paths as used to be the case and in addition it handles adding a blueprint. I.e. you can send 2 types of data to this one endpoint, either `["path","path"]` or `{"id": "id"}`.

PR #201, opened 2015-12-02
HWKINVENT-162 - Relationships endpoint not working for resources because of ambiguity of the URLs

jkremser:
- changing the `stuff/relationships` into `relationshipsOf/{canonicalPath}`
- adding some i-tests
- fixing NPE in `EntityNotFoundExceptionMapper` (unrelated to the $subject)

### DIFF
--- a/hawkular-inventory-rest-api/src/main/java/org/hawkular/inventory/rest/RestRelationships.java
+++ b/hawkular-inventory-rest-api/src/main/java/org/hawkular/inventory/rest/RestRelationships.java
@@ -32,6 +32,7 @@ import javax.inject.Inject;
 import javax.ws.rs.Consumes;
 import javax.ws.rs.DELETE;
 import javax.ws.rs.DefaultValue;
+import javax.ws.rs.Encoded;
 import javax.ws.rs.GET;
 import javax.ws.rs.POST;
 import javax.ws.rs.PUT;
@@ -63,7 +64,6 @@ import org.hawkular.inventory.api.paging.Page;
 import org.hawkular.inventory.api.paging.Pager;
 import org.hawkular.inventory.rest.json.ApiError;
 import org.hawkular.inventory.rest.json.JsonLd;
-import org.hawkular.inventory.rest.security.EntityIdUtils;
 
 import com.fasterxml.jackson.core.JsonProcessingException;
 import com.fasterxml.jackson.databind.ObjectMapper;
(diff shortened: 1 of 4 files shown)


metlos:
Not sure I like the `relationshipsOf/...`. While it's grammatically correct I'd just prefer simple `relationships/...` in URLs. If we decide to go to the prefix URLs for other things, we'd have `resources/<<path>>`, `children/<<path>>`, etc. instead of  `resourcesOf/<<path>>`, `childrenOf/<<path>>`..

jkremser:
`childrenOf`, `parentOf`, `relationshipsOf`, etc. are different types of prefixes than `resources`, `metrics`, `resourceTypes`, etc.

The problem with "`s/relationshipsOf/relationship/g`" is that there is already endpoint for `/relationships/{relationshipId}` and it expects the relationshipId as a path param (this can be changed to canonical path of the relationship, but the relationship ids are unique across all the inventory so this was easier)

jkremser:
`children`, `parent`, `parents`, etc. at least returns the same entity type as the parameter. This doesn't hold for queries like `resourceType/foo/resources`

We can establish some consistent strategy like `blabla/inventory/whatIwant/canonicalPathToIt` and overloading the behavior based on the type of the `canonicalPath` for instance for the query mentoined before (`resourceType/foo/resources`) it may work in a way:

`blabla/inventory/resources/CPathToTheResourceType` and we may expect that user wants all the resources of this resource type

same for metrics of a resource:
`blabla/inventory/metrics/CPathToTheResource` -> get all the metrics associated to this resource

... and other 1:N relationships (actually M:N can work this way as well, but it's more difficult to implement)

in a way, same for the relationships:
`blabla/inventory/relationships/CPathToTheRelationship` -> return the relation
`blabla/inventory/relationships/CPathToTheNonRelationship` -> return all the relations of the entity

not sure how to handle the children/parent, etc. 
perhaps something like
`blabla/resources/children/{CPath}`, but what if the CPath points to the metric/resourcetype, etc. do we want to fail here, or provide some smartness like give me all the children resources of given type

hawkular-inventory-rest-api/src/main/java/org/hawkular/inventory/rest/RestRelationships.java
```diff
@@ -164,64 +164,102 @@ public Response getRelationship(@PathParam("relationshipId") String relationship
...
         ResolvableToSingleWithRelationships<Relationship, Relationship.Update> resolvable =
-                getResolvableFromCanonicalPath(cPath);
+                (ResolvableToSingleWithRelationships<Relationship, Relationship.Update>) inventory.inspect(cPath,
+                        ResolvableToSingleWithRelationships.class);
 
         Relationships.Direction directed;
         CanonicalPath theOtherSide;
-        String[] chunks = path.split("/");
-        String currentEntityId = chunks[chunks.length - 1];
+        String currentEntityId = cPath.getSegment().getElementId();
```
pavolloffay, line 262: :+1:  this solves a bug I just found

pavolloffay, line 262: This is blocker for me, It is not possible to create relationship from tenant to metric/metricType.

pavolloffay:
What about moving whole relationships under /inventory/path/relationships

There is already endpoint for getting relationships 
https://github.com/hawkular/hawkular-inventory/blob/master/hawkular-inventory-rest-api/src/main/java/org/hawkular/inventory/rest/RestPath.java#L93

EDIT: 

path  /path/e;test/r;res1/res2/relationships  would match /path/<.*>/relationships which is wrong

jkremser:
Closing as this is out dated

PR #202, opened 2015-12-04
[HWKINVENT-153] Detection of identical resource and metric types across feeds and tenant.

metlos:
This is done using an internal "IdentityHash" vertex that is referenced
from the tenant (to allow for universal access) using
"__containsIdentityHash" edges (so that it doesn't "pollute" the
potentially crowded "contains" edges.

These identity hash vertices are then referenced by the individual resource
and metric types using the "__identityHash" edges. The identity hash
of metric and resource types are recomputed on any update of their
"hash-significant" values or of any values of their contained entities
(op types, data).

To query for identical resource types, then, is a simple matter of going
out to the hash vertex and then back across all the "__identityHash"
edges.

The identity hash vertices are cleared out automatically when they are no
longer referenced by any resource/metric type.


jkremser:
Could you please rebase / merge the master to it?

metlos:
@Jiri-Kremser finally I got this to build successfully on the latest master.

PR #204, opened 2015-12-08
HWKINVENT-163

pavolloffay:
Expose inventory query capabilities on the bus so other components can query inventory entities through the bus.


hawkular-inventory-api/src/main/java/org/hawkular/inventory/api/model/Tenant.java
```diff
@@ -53,6 +53,7 @@ public Tenant(CanonicalPath path, Map<String, Object> properties) {
 
     public Tenant(String name, CanonicalPath path, Map<String, Object> properties) {
         super(name, path, properties);
+        getId();
```
metlos: What is the purpose of this?

hawkular-inventory-api/src/main/java/org/hawkular/inventory/api/Inventory.java
```diff
@@ -505,4 +506,8 @@ public Single visitUnknown(Void parameter) {
                                                                   String... relationshipNames);
 
     Configuration getConfiguration();
+
+    default <T extends AbstractElement> List<T> execute(Query query, Class<T> requestedEntity, Pager pager) {
```
metlos: Shouldn't this return a `Page<T>` instead of a `List<T>`?

hawkular-inventory-bus-api/src/main/java/org/hawkular/inventory/bus/api/InventoryAbstractMessage.java
```diff
@@ -0,0 +1,53 @@
...
+        mapper.addMixIn(CanonicalPath.class, CanonicalPathMixin.class);
+        return mapper;
+    }
+
+    public static ObjectMapper buildObjectMapperForDeserialization() {
+        /**
+         * @see url(https://stackoverflow.com/questions/19694928/
+         * jackson-jersey-deserialize-exception-for-id-type-id-class-no-such-class)
+         */
+        Thread.currentThread().setContextClassLoader(InventoryAbstractMessage.class.getClassLoader());
```
metlos: Does this need to be a perpetual change or is it bound by this method? If it is the latter, I'd reset the context classloader to its original value in a finally block in this method.

hawkular-inventory-itest-parent/hawkular-inventory-itest/src/main/groovy/org/hawkular/inventory/rest/test/InventoryITest.groovy
```diff
@@ -22,9 +22,7 @@ import org.hawkular.inventory.api.model.Resource
 import org.junit.*
 
 import static org.hawkular.inventory.api.Relationships.WellKnown.*
-import static org.junit.Assert.assertEquals
-import static org.junit.Assert.assertTrue
-import static org.junit.Assert.fail
+import static org.junit.Assert.*
```
metlos: Are wildcard imports allowed by our formatting guidelines, @Jiri-Kremser ?

hawkular-inventory-bus-api/src/main/java/org/hawkular/inventory/bus/api/InventoryAbstractMessage.java
pavolloffay: If it is set back to the original class loader deserialization fails http://pastebin.com/CHcqcysC

hawkular-inventory-itest-parent/hawkular-inventory-itest/src/main/groovy/org/hawkular/inventory/rest/test/InventoryITest.groovy
jkremser: Checkstyle is ok with that, so I guess it's ok

hawkular-inventory-bus-api/src/main/java/org/hawkular/inventory/bus/api/InventoryAbstractMessage.java
metlos: Ok, but in the EE container environment, changing the context classloader is considered dangerous.

So what if, instead you returned a custom `ObjectMapper` decorator that would set and reset the context classloader in every method? There are only like 300 of them :wink:.

Or maybe even better - this should be fixed in bus that should set and reset the context classloader before and after using this object mapper, as per https://stackoverflow.com/questions/19694928/+jackson-jersey-deserialize-exception-for-id-type-id-class-no-such-class.

pavolloffay: Ok It works fine with change in bus :).

hawkular-inventory-api/src/main/java/org/hawkular/inventory/api/Inventory.java
pavolloffay: :+1: done

hawkular-inventory-bus-api/src/main/java/org/hawkular/inventory/bus/api/InventoryAbstractMessage.java
pavolloffay: https://github.com/hawkular/hawkular-bus/pull/55/files

metlos:
I believe this needs more work in a couple of areas:
- the `Query` class needs to be polished, because the API is not that good for public use
- `equals()`, `hashCode()` are not implemented in all filters and query fragments
- most importantly I am not sure if JNDI wouldn't be a better path forward with RPC type of access - using async bus for querying doesn't seem 100% right to me. That said, if JNDI doesn't work out (which it should), then we can move forward with this.

metlos:
Just to track progress - we agreed with @pavolloffay that he'll take a look at the `Query` class and pimp up its interface a little bit. After that this will be mergeable.

hawkular-inventory-api/src/test/java/org/hawkular/inventory/api/test/QueryTest.java
```diff
@@ -113,4 +116,34 @@ public void testOptimizationRessurectsAfterNonCanonicalSection() throws Exceptio
...
+                .branch().with(qMetricsRelationships).done()
+                .branch().with(qTypesRelationships).done().build();
+
+        // new query
+        Query.Builder builder = new Query.Builder().path().with(new NoopFilter()).builder()
+                .branch().path().with(With.path(cp1), SwitchElementType.incomingRelationships(),
+                        RelationWith.name("name")).builder().done()
+                .branch().path().with(With.path(cp2), SwitchElementType.incomingRelationships(),
+                        RelationWith.name("name")).builder().done();
+
```
pavolloffay, line 144: @metlos example of using Builder. Is it ok?

metlos, line 144: Not sure I like the `.builder().branch()` and `.builder().done()` "bridges". Why not `.branch()` and `.done()` directly?

PR #206, opened 2015-12-11
HWKINVENT-167

jkremser:
upgrading accounts to `1.1.3.Final`, it's fixed in this version

..also removing the `@CurentUser` annotation and using the `@Inject Instance<Persona>` instead of `@Inject Persona` which is the suggested way.


jpkrohling:
Looks good to me :)

PR #208, opened 2015-12-16
Use java:/ConnectionFactory name to get the JMX connection factory

ppalaga:


ppalaga:
@metlos this is an attempt to fix https://github.com/hawkular/hawkular-agent/pull/127 but it does not help as much as I hoped. Could you please have a look what is wrong with https://github.com/hawkular/hawkular-agent/pull/127 ?

PR #210, opened 2015-12-17
Increase the wait time for cassandra to start up.

metlos:


metlos:
This is obviously non-optimal solution until we have something more elaborate like resource adapter or something that we could wait on.

PR #211, opened 2015-12-17
Upgrade to Accounts 2.0.9.Final and Commons 0.3.1.Final

ppalaga:


ppalaga:
I need this in Command Gateway and Agent. I hope I'll cause no damage by not waiting for someone else's review. cc @metlos

PR #212, opened 2015-12-18
HWKINVENT-170 Rest+WS Events

jkremser:
Exposing the inventory events via REST endpoint and via WebSocket endpoint
for the REST endpoint, it holds the connection for 20 seconds and respond with all the events that happened in this time window.

As for the WS endpoint, it's more natural, every time the event occurs it's directly forwarded to the connected client in JSON form.

~~Please, don't merge yet, it needs more love.~~

### DIFF
--- a/hawkular-inventory-bus-api/src/main/java/org/hawkular/inventory/bus/api/DataEntityEvent.java
+++ b/hawkular-inventory-bus-api/src/main/java/org/hawkular/inventory/bus/api/DataEntityEvent.java
@@ -20,6 +20,7 @@ import java.util.Map;
 
 import org.hawkular.inventory.api.Action;
 import org.hawkular.inventory.api.model.DataEntity;
+import org.hawkular.inventory.api.model.Tenant;
 
 /**
  * @author Lukas Krejci
@@ -29,8 +30,8 @@ public final class DataEntityEvent extends InventoryEvent<DataEntity> {
     public DataEntityEvent() {
     }
 
-    public DataEntityEvent(Action.Enumerated action, DataEntity object) {
-        super(action, object);
+    public DataEntityEvent(Action.Enumerated action, Tenant tenant, DataEntity object) {
+        super(action, tenant, object);
     }
 
     @Override
--- a/hawkular-inventory-bus-api/src/main/java/org/hawkular/inventory/bus/api/InventoryEvent.java
+++ b/hawkular-inventory-bus-api/src/main/java/org/hawkular/inventory/bus/api/InventoryEvent.java
@@ -48,6 +48,7 @@ import com.fasterxml.jackson.databind.ObjectMapper;
 public abstract class InventoryEvent<T extends AbstractElement<?, ?>> extends AbstractMessage {
 
     private Action.Enumerated action;
+    private Tenant tenant;
     private T object;
 
     public static Class<? extends InventoryEvent<?>> determineEventType(Message message) {
@@ -96,7 +97,7 @@ public abstract class InventoryEvent<T extends AbstractElement<?, ?>> extends Ab
         }
     }
 
-    public static InventoryEvent<?> from(Action<?, ?> action, Object object) {
+    public static InventoryEvent<?> from(Action<?, ?> action, Tenant tenant, Object object) {
         if (object == null) {
             throw new IllegalArgumentException("object == null");
         }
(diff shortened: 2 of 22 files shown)


hawkular-inventory-bus-api/src/main/java/org/hawkular/inventory/bus/api/DataEntityEvent.java
```diff
@@ -29,8 +30,8 @@
     public DataEntityEvent() {
     }
 
-    public DataEntityEvent(Action.Enumerated action, DataEntity object) {
-        super(action, object);
+    public DataEntityEvent(Action.Enumerated action, Tenant tenant, DataEntity object) {
```
metlos, line 33: What do you need the tenant for? I haven't spotted anything in this PR that would require it...

hawkular-inventory-bus-api/src/main/java/org/hawkular/inventory/bus/api/InventoryEvent.java
```diff
@@ -154,6 +156,14 @@ public void setAction(Action.Enumerated action) {
         this.action = action;
     }
 
+    public Tenant getTenant() {
+        return tenant;
+    }
+
+    public void setTenant(Tenant tenant) {
```
metlos, line 163: Is this used? me don't like ugly mutable state ;)

hawkular-inventory-bus/src/test/java/org/hawkular/inventory/bus/BusTest.java
```diff
@@ -228,9 +238,11 @@ public void sendMetricEvent() throws Exception {
 //        broker.start();
```
metlos, line 238: This used to work - either explain why the test is no longer needed or make it work again, pls.

hawkular-inventory-rest-api/src/main/java/org/hawkular/inventory/rest/RestApiLogger.java
```diff
@@ -28,7 +28,7 @@
 /**
  * Logger definitions for Jboss Logging for the rest api
  *
- * Code range is 2000-2999
+ * Code range is 2000-2899
```
metlos, line 31: Why this decrease? What do you intend 2900-2999 for?

metlos, line 31: Ah, I see :) Didn't scroll down enough :)

hawkular-inventory-rest-api/src/main/java/org/hawkular/inventory/websocket/WebsocketEvents.java
```diff
@@ -0,0 +1,187 @@
...
+    }
+
+    @OnError
+    public void onError(Throwable error) {
+        WebsocketApiLogger.LOGGER.errorHappened(error);
+    }
+
+    @OnMessage
+    public void handleMessage(String message, Session session) {
+        WebsocketApiLogger.LOGGER.onMessage(session.getId(), message);
```
metlos, line 132: Shouldn't we also output something to the caller?

hawkular-inventory-bus/src/test/java/org/hawkular/inventory/bus/BusTest.java
jkremser, line 238: > This used to work - either explain why the test is no longer needed or make it work again, pls.

Not sure I follow. Are you referring to the `//        broker.start();` I didn't touch that.

hawkular-inventory-bus-api/src/main/java/org/hawkular/inventory/bus/api/DataEntityEvent.java
jkremser, line 33: We don't want to expose events belonging to a different tenant (security issues)

Right, it's not used. I was assuming that `inventory.observable(interest)` would provide an `Observable<? extends InventoryEvent>` so that I could filter the events of a given tenants but this was not the case.

I guess the InventoryEvent is used only for the bus communication. However, it may be convenient to have the tenant id in the json payload, instead of parsing the cpaths, wdyt?

At the end of the day I used this http://git.io/vuzrI and http://git.io/vuzrs for filtering by tenant id.

I don't like how the `RestEvents.getFilter()` method looks like, because if a new action is added, this place needs to be updated.. Other option would be to have these filters (String tenantId -> boolean) inside the `Action` itself.

hawkular-inventory-bus-api/src/main/java/org/hawkular/inventory/bus/api/InventoryEvent.java
jkremser, line 163: Me neither, I just wanted to be consistent with the rest of the class, because there is also `{get|set}Object()`.

I can remove the setters for tenant and object. They are not being used anyway.

PR #213, opened 2015-12-21
[HWKINVENT-168] no duplicate relationships + a couple of other fixes

metlos:
This contains a couple of things:
- fix for deleting data nodes of the data entities that were left behind during recursive delete
- fix for finding relationships by canonical path
- refactoring of the build so that the inventory unit tests run on both tinkergraph (quick) and Titan (distribution-compatible). This will make sure that we're testing with the same stuff with deploying with.
- finally, fix for https://issues.jboss.org/browse/HWKINVENT-168 itself - this was caused by the `linkWith()` method from generic relationships using custom code for creating them. This has now been modified to use the common code path with the rest of the codebase to ensure uniform behavior.

### DIFF
--- a/hawkular-inventory-api/src/main/java/org/hawkular/inventory/base/Util.java
+++ b/hawkular-inventory-api/src/main/java/org/hawkular/inventory/base/Util.java
@@ -30,6 +30,7 @@ import java.util.HashSet;
 import java.util.Random;
 import java.util.Set;
 import java.util.function.BiConsumer;
+import java.util.function.Consumer;
 import java.util.stream.Collectors;
 
 import org.hawkular.inventory.api.Action;
--- a/hawkular-inventory-api/src/main/java/org/hawkular/inventory/base/Util.java
+++ b/hawkular-inventory-api/src/main/java/org/hawkular/inventory/base/Util.java
@@ -24,6 +24,7 @@ import static org.hawkular.inventory.api.Relationships.Direction.both;
 import static org.hawkular.inventory.api.Relationships.Direction.outgoing;
 import static org.hawkular.inventory.api.Relationships.WellKnown.contains;
 import static org.hawkular.inventory.api.Relationships.WellKnown.defines;
+import static org.hawkular.inventory.api.Relationships.WellKnown.hasData;
 
 import java.util.HashSet;
 import java.util.Random;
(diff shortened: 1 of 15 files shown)


hawkular-inventory-api/src/main/java/org/hawkular/inventory/base/Util.java
```diff
@@ -352,14 +356,15 @@ public static Query extendTo(TraversalContext<?, ?> context, Path path) {
...
-            } else {
-                deleted.add(entity);
-            }
-            deletedRels.addAll(context.backend.getRelationships(entity, both));
+                context.backend.getRelationships(e, outgoing, hasData.name()).forEach(rel -> {
+                    dataToBeDeleted.add(context.backend.getRelationshipTarget(rel));
+                });
+            };
+
+            categorizer.accept(entity);
```
jkremser, line 370: is this line needed here?

hawkular-inventory-impl-tinkerpop-parent/hawkular-inventory-impl-tinkerpop-titan-provider/src/test/resources/cassandra-config.yaml
```diff
@@ -0,0 +1,106 @@
...
+tombstone_failure_threshold: 100000
+tombstone_warn_threshold: 1000
+trickle_fsync: false
+trickle_fsync_interval_in_kb: 10240
+truncate_request_timeout_in_ms: 60000
+write_request_timeout_in_ms: 2000
+
+commitlog_directory: ${project.build.directory}/cassandra/commitlog
+saved_caches_directory: ${project.build.directory}/cassandra/saved_caches
+data_file_directories: [${project.build.directory}/cassandra/data]
```
jkremser, line 106: This file may diverge from http://git.io/vEZnq in the future, but I have no better solution.

Perhaps some Maven magic.

metlos, line 106: IMHO that doesn't matter that much. All we care about is bringing Titan up with C*. It shouldn't matter too much for the itests if C* is configured differently.

hawkular-inventory-api/src/main/java/org/hawkular/inventory/base/Util.java
metlos, line 370: Yes, so that the `entity` ends up in the `deleted` or `verticesToDeleteThatDefineSomething` sets.

PR #220, opened 2016-01-25
load-tests: catching up on the rest api + running them by travis

jkremser:
This PR:
- makes the load tests part of the travis build (by default it runs only the minimal configuration, just to test it works). No conclusions should be made upon these perf results, because the travis CI performance is very volatile.
- adds a new scenario for inserting (`n` times) all the entities that are normally reported by agent for 1 hawkular server with 1 running WF -> the structure among the data is closer to reality that the previous scenario. All the entities are stored in the `data.json` in which the `${feedId}` is parametrized, so what the test does is, creating these same entities for `n` different feeds.
- makes some refactoring so it's not a big ball of mud anymore


Jiri-Kremser:
WIP (from "salina" ;)

metlos:
We can now run inventory with the postgres backend - see https://github.com/metlos/hawkular-inventory/commit/8b068506bb4c1d76480296646e29c094fb600da0 (this is in an experimental branch).

PR #221, opened 2016-01-26
Reinstated the Secret Store

jpkrohling:
This kinda reverts 1eee1e5 : the dependency has to be declared here, so that it's included in the WAR, but the version can be managed somewhere else.


jpkrohling:
@metlos , @Jiri-Kremser , could you please review/merge?

PR #224, opened 2016-02-05
Introduce Inventory feature pack to ease the installation of inventory into Wildfly.

metlos:
Also, in a separate commit, itests updated to take advantage of the feature pack.


metlos:
@ppalaga would you please have a look at this? travis is acting up, this builds just fine locally.

ppalaga:
The FP plugin fails if the `src/main/resources/modules` directory is not there in the FP module. Just run `echo "wildfly-feature-pack-plugin fails if src/main/resources/modules does not exist" > hawkular-inventory-feature-pack/src/main/resources/modules/README.adoc` from to top inventory folder to make git aware of the directory.

ppalaga:
With the named changes, I can build locally, and the itests pass. The first review round is thus finished.

metlos:
@ppalaga - updated according to your comments.. Would you please review again?

PR #225, opened 2016-02-10
Fix the parent artifactId in itest module

ppalaga:


ppalaga:
@metlos would you please review and eventually merge? BTW, this might fix rev-api to work on Travis again.

metlos:
Oops, good catch.

PR #229, opened 2016-02-19
swager update

pavolloffay:
Update to parent 35 broke docgen profile. For generating adoc is used groovy script from Metrics. It would be good to move it to hawkular-build-tools where lives old mustache template. 

Note: 
`<version.com.github.kongchen.swagger-maven-plugin>3.1.1</version.com.github.kongchen.swagger-maven-plugin>`
overrides parent's 3.1.2 which produces NPE instead of json...

### DIFF
--- a/hawkular-inventory-rest-api/pom.xml
+++ b/hawkular-inventory-rest-api/pom.xml
@@ -232,26 +232,32 @@
     </dependency>
 
     <!-- docs -->
-
     <dependency>
-      <groupId>com.wordnik</groupId>
+      <groupId>io.swagger</groupId>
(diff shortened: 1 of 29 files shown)


pavolloffay:
@Jiri-Kremser could you please review?

pavolloffay:
It passes locally. 

The log length has exceeded the limit of 4 Megabytes (this usually means that test suite is raising the same exception over and over).

The build has been terminated.

pavolloffay:
I don't know why the build was failing, I had to restart it couple of times. Locally it passes without problems.

PR #231, opened 2016-02-25
Simplifying the Revapi config very much.

metlos:
Now if an incompatible change occurs, all you need to do is to
mvn package revapi:update-versions
and it will change the version of the artifact so that the changes made
and the semver version are in accord.


pom.xml
```diff
@@ -234,6 +231,11 @@
                 <![CDATA[
                   {
                     "revapi" : {
+                      "semver": {
+                        "ignore": {
+                          "enabled": true
```
jkremser, line 244: nice, does it mean the plugin will automatically increase the minor version if there is an incompatible change? (because we are < 0.x.y)

Why the descriptions are not needed anymore? Is it because of the `< 0.x.y` thing?

metlos, line 244: if there is an incompatible change, it will fail the build and you can remedy it by running `mvn package revapi:update-versions` which will automatically change the version in the pom.

So in another words yes and yes ;)

metlos, line 244: As for why it fails the build the first time - I don't think there is a way in maven to do it in go - essentially revapi needs the build to build the artifact before it can compare it with the old version at which point it is too late to reset the version in the pom.

metlos:
This passes locally. Ignoring Travis..

PR #232, opened 2016-02-29
Take the subsystems.xml and template.xml unchanged or transformed from the depen

ppalaga:


ppalaga:

ppalaga:
@metlos strange enough, it is passing here locally running with `-Pitest,ltest`. Do you have any idea what is Travis not liking about this PR?

ppalaga:
@metlos It is passing now. I wonder if anything relevant has changed in master?

metlos:
Yes, #233 has been merged, which made some corrections that made itests and load tests much less noisy (and the impl more correct).

PR #236, opened 2016-03-01
Bulk insert - duplicates

jkremser:
Making the bulk insert more robust so that it can deal with duplicated elements.
This PR also fixes the load tests that has some duplicates in the "recorded" json from agent.

_With this change, the Hawkular is happy again. Without it, there is nothing in the inventory, except the feed, because whole bulk insert is rolled back._

Note: agent is still sending duplicate elements, but inventory is now immune to it.


metlos:
I just built this PR locally using the same args as Travis and it built fine. Not sure what Travis' problem is that it is constantly failing the load tests with "connection refused".

PR #243, opened 2016-04-14
Use docgen groovy script from build-tools

pavolloffay:
Use docgen groovy script from hawkular-build-tools.


pavolloffay:
Locally it is passing with no problems.

PR #244, opened 2016-04-14
[HWKINVENT-177] Make the created entity caching optional.

metlos:
And update the docs to warn about the possibility of seeing stale data.


metlos:
```
mvn clean install -Pitest
```

is passing locally for me.

jkremser:
wow, green build :+1:

PR #245, opened 2016-04-15
Upgr to hawkular-parent 41 (Cassandra 3.5 and version.org.wildfly.bom)

ppalaga:


ppalaga:
@metlos @Jiri-Kremser this is using Embedded Cassandra with configuration targeted at low memory usage. Let's see if Travis likes it.

metlos:
This is passing locally, even though Travis doesn't like it. @ppalaga, should I merge it or do you want to do another round of experiments?

ppalaga:
@metlos this is again failing in impl-tinkerpop-titan-provider. Do you have any explanation of that? 

I think that merging this will not make the present state worse, hence I vote for merging.

Independently of that, I'll try to test this commit in a branch where your recent PRs are not present to see it makes the itests pass there at least.

metlos:
This passed for me with the latest master.

PR #241, opened 2016-04-13
Transform the itests from Groovy to Java

ppalaga:

PR #246, opened 2016-04-18
Restore the modifications of the logging subsystem by inv feature pack.

metlos:
Additionally, up the commons version to be on par with accounts.


metlos:
This got merged together with #241 that depended on this functionality.

PR #247, opened 2016-04-19
Make the Cassandra config in Titan provider as minimalistic as possible

ppalaga:


ppalaga:
@metlos @Jiri-Kremser I am curious if this is going to pass.

ppalaga:
Looks like a slight improvement against the current master: It is not failing in Titan Provider anymore, but in itests (where the container is not able to boot). Let me try to restart if the result is the same...

ppalaga:
@metlos switching to Ubuntu Trusty VM with 7.5 GB RAM helps, but the build takes 28 min, because caching of resources is not available for this VM type - see https://docs.travis-ci.com/user/ci-environment/#Virtualization-environments . But this is still better than fail for every build. WDYT?

PR #249, opened 2016-04-21
[HWKINVENT-171] Rest docs improvements

metlos:
I finally found a way of persuading Swagger to distinguish between the various Blueprint and Update classes we have in our model.


metlos:
@ppalaga, could you take a quick look at this? I'd like to have these improvements in `0.15.0.Final` so that the docs gradually suck less and less ;)

hawkular-inventory-api/src/main/java/org/hawkular/inventory/api/model/StructuredData.java
```diff
@@ -41,10 +44,12 @@
  * @author Lukas Krejci
  * @since 0.3.0
  */
+@ApiModel(value = "JSON", description = "Just free form JSON.")
```
ppalaga: `s/Just free form JSON./Just a free form JSON./`

PR #250, opened 2016-04-21
Another go at minimizing the output of the travis build.

metlos:
The only downside to this that I can think of is that this will not print out the stacktraces.
But it will output all the warns and errors so at least that will be visible.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -31,10 +31,30 @@ install:
 - git fetch origin --unshallow
 
 before_script:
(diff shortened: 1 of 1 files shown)


metlos:
Why the hell is the output different from what I see locally in my travis docker image?

jkremser:
`No output has been received in the last 10m0s, this potentially indicates a stalled build or something wrong with the build itself.`

The titan tests probably

PR #251, opened 2016-04-21
Filtering the output of build when building on Travis

jkremser:
ouch, too late #250 was merged 10 mins ago

Well, just in case you want to have the full control of the output (perhaps in the future or if the current approach wont prove itself to work well)

I am raising this PR not to lose the code.


metlos:
Thanks Jirka, for sharing this. I'll close this PR so that it doesn't take up place in the PR queue, but we can resurrect it in the future.

PR #255, opened 2016-04-27
Docs fix

jkremser:
the reason the docs werent deployed is that `PROJECT_VERSION` property was empty, probably because of the mvn opts.

https://travis-ci.org/hawkular/hawkular-inventory/builds/126100490#L1140


jkremser:
it worked, the docs are deploying
but the snapshot deployment to the nexus is not working. 

https://travis-ci.org/hawkular/hawkular-inventory/builds/126152012#L1230 :
`Not authorized , ReasonPhrase:Unauthorized.`

it takes the credentials from here: https://git.io/vwrtZ
and these properties are encoded as secrets in the `.travis.yml`, iirc @stefannegrea was doing this, do we still need it?

PR #258, opened 2016-05-09
[HWKINVENT-165] Move SwitchElementType to API

metlos:
and properly integrate it into query optimizer and gremlin query construction logic.


jkremser:
@pavolloffay could you look, please? I don't have context.

hawkular-inventory-impl-tinkerpop-parent/hawkular-inventory-impl-tinkerpop/src/main/java/org/hawkular/inventory/impl/tinkerpop/FilterVisitor.java
```diff
@@ -242,26 +242,39 @@ public void visit(HawkularPipeline<?, ?> query, SwitchElementType filter, QueryT
...
+                    state.setComingFrom(null);
                     query.bothV();
                 } else {
+                    state.setInEdges(true);
+                    state.setComingFrom(Direction.BOTH);
                     query.bothE();
                 }
                 break;
         }
+        state.setExplicitChange(true);
```
pavolloffay, line 277: This `state#ExplicitChange` is used in  `static <S, E> void finishPipeline`, but I'm wondering why it always sets the `ExplicitChange(true);` to true.

metlos, line 277: The query rewriting takes advantage of the fact that some of the node properties are mirrored on the edges. When constructing the gremlin query, the translation state tracks whether the query is currently on an edge or an entity. When a "piece of work" is finished during the translation, the `finishPipeline` method is called that wraps up that piece. It needs to go back to the same state (vertices/edge) where it came from - usually this is vertices, because the vast majority of the filters work on entities.

When we made `SwitchElementType` filter public, we added the ability to explicitly change the type of element (vertex/edge) we're querying during the course of the pipeline. The `finishPipeline` method must not "reset" such explicit changes.

So whatever the state change is induced when applying the type-switching filter, it is marked as explicit because it comes from a filter, not from query translation internal optimization logic.

What I am not 100% sure about is the resetting of the flag to false in `QueryTranslationState#setInEdges` et al. I'll have to think about whether that is necessary or not.

PR #259, opened 2016-05-12
HWKINVENT-181 Introduce Security SPI

ppalaga:


hawkular-inventory-api/src/main/java/org/hawkular/inventory/api/Inventory.java
```diff
@@ -101,6 +110,91 @@
...
+                    ResourceType.class,
+                    MetricType.class,
+                    Resource.class,
+                    Metric.class,
+                    OperationType.class,
+                    MetadataPack.class)
+            .stream().collect(Collectors.toMap(Class::getName, Function.identity()));
+            entityTypes = Collections.unmodifiableMap(m);
+        }
+        static String shortenIfNeeded(CanonicalPath path) {
```
metlos: This has no place in the API.. This method was introduced to integrate with hwk accounts that didn't like our possibly long paths that are the only thing uniquely identifying an entity. Inventory always stores the full path and I don't think we need to introduce yet another unique identifier for an entity in inventory itself (i.e. having 2 ways of identifying an entity).

hawkular-inventory-api/src/main/java/org/hawkular/inventory/api/Inventory.java
```diff
@@ -101,6 +110,91 @@
...
+            }
+        }
+    }
+    static String getStableId(CanonicalPath path) {
+        return InventotyUtils.shortenIfNeeded(path);
+    }
+    static String getStableId(AbstractElement<?, ?> element) {
+        return getStableId(element.getPath());
+    }
+    static CanonicalPath toCanonicalPath(String restPath) {
```
metlos: This is coupling the API to a particular REST API implementation (that will change with my next PR ;) ).

hawkular-inventory-rest-api/pom.xml
```diff
@@ -59,6 +59,18 @@
     </dependency>
 
     <dependency>
+      <groupId>org.hawkular.inventory</groupId>
+      <artifactId>hawkular-inventory-security-spi</artifactId>
+      <version>${project.version}</version>
+    </dependency>
+
+    <dependency>
```
metlos, line 67: Maybe this could be moved to the `hawkular-inventory-dist` which assembles the full inventory war?

ppalaga:
@metlos this does not remove Accounts, It just introduces the SPI with the two planned implementations.

Open problems:
- I throw an `IllegalStateException` from the `HttpHeaderTenantIdProducer` if the header is not set https://github.com/hawkular/hawkular-inventory/pull/259/commits/d6ace572ffd55405dd9d079ec96f6fa0f9d5e9d6#diff-8563bf0980d454e58e40a903d0fd4836R50 That makes the request return 500, which I  admit is not optimal. Can you see a way to return a better code with a better message?
- The original Accounts-based `InventorySecurity` used to autocreate the tenant.  https://github.com/hawkular/hawkular-inventory/pull/259/commits/d6ace572ffd55405dd9d079ec96f6fa0f9d5e9d6#diff-e496894d9c6dba4ce79ea79691d9bfbcR126 which I had to comment out because I saw no way to inject an `inventory` bean to `InventorySecurity`. Should the tenant auto-creation not be solved elsewhere - e.g. in the REST API, e.g. in a ServletFiltet (that could possible be made inactive using e.g. a System property)?

hawkular-inventory-rest-api/pom.xml
ppalaga, line 67: No, rest API has a real compile time dependency on `hawkular-inventory-security-spi` and therefore the dependency on `hawkular-inventory-security-spi` cannot be moved to `hawkular-inventory-dist`

hawkular-inventory-api/src/main/java/org/hawkular/inventory/api/Inventory.java
ppalaga: There is a reference to `getStableId()` outside the security package: https://github.com/hawkular/hawkular-inventory/pull/259/files#diff-0646e1614dd0f2d2c68141bcd5bc6fdaR107 which is why I moved it here where both REST API and Security implementations can depend on it.

ppalaga: I am OK with canning this to Accounts-based Security just tell me what should happen with https://github.com/hawkular/hawkular-inventory/pull/259/files#diff-0646e1614dd0f2d2c68141bcd5bc6fdaR107 ?

hawkular-inventory-api/src/main/java/org/hawkular/inventory/api/Inventory.java
ppalaga: Yes, toCanonicalPath can go back to REST API. Any preference which class I should put it into?

hawkular-inventory-api/src/main/java/org/hawkular/inventory/api/Inventory.java
```diff
@@ -101,6 +110,91 @@
...
+                    Tenant.class,
+                    Environment.class,
+                    Feed.class,
+                    ResourceType.class,
+                    MetricType.class,
+                    Resource.class,
+                    Metric.class,
+                    OperationType.class,
+                    MetadataPack.class)
+            .stream().collect(Collectors.toMap(Class::getName, Function.identity()));
```
jkremser:
```
<nitpicker>
Arrays.asList('a','b','c').stream() -> Stream.of('a', 'b', 'c')
</nitpicker>
```

hawkular-inventory-rest-api/src/main/java/org/hawkular/inventory/rest/json/RelationshipEmbeddedJacksonSerializer.java
```diff
@@ -104,7 +104,7 @@ private void serializeEntity(CanonicalPath entity, JsonGenerator jsonGenerator,
         jsonGenerator.writeStartObject();
 
         jsonGenerator.writeFieldName(FIELD_ID);
-        jsonGenerator.writeString(EntityIdUtils.getStableId(entity));
+        jsonGenerator.writeString(Inventory.getStableId(entity));
```
jkremser: Returning `entity.toString()` could be ok. iirc, the stable id should be the same as the canonical path except for the very long canonical paths.

jkremser:
> I throw an IllegalStateException from the HttpHeaderTenantIdProducer if the header is not set

What about just setting the tenant to "hawkular" to stay backward compatible?

hawkular-inventory-api/src/main/java/org/hawkular/inventory/api/Inventory.java
ppalaga: Thanks for a nice tip, @Jiri-Kremser !

ppalaga:
> > I throw an IllegalStateException from the HttpHeaderTenantIdProducer if the header is not set
> 
> What about just setting the tenant to "hawkular" to stay backward compatible?

Have we ever set "hawkular" as a tenantId header value in the past?

hawkular-inventory-rest-api/src/main/java/org/hawkular/inventory/rest/json/RelationshipEmbeddedJacksonSerializer.java
metlos: This actually should go away, because an entity cannot be found using the stable id - stable id is merely an identifier used to find the security details of the entity, but NOT the entity itself in inventory. The stable id is not stored in inventory and the full canonical path is not recoverable from the stable id.

hawkular-inventory-api/src/main/java/org/hawkular/inventory/api/Inventory.java
metlos: I commented on that line. IMHO, it has been incorrect the whole time and should be replaced with CanonicalPath.toString().

hawkular-inventory-api/src/main/java/org/hawkular/inventory/api/Inventory.java
metlos: None in particular, because it will disappear anyway. The new REST API will not use it.

hawkular-inventory-rest-api/pom.xml
metlos, line 67: I was talking about the hard compile dependency on `hawkular-inventory-security-permissive`. The dependency on the security SPI is of course alright.

metlos:
> > I throw an IllegalStateException from the HttpHeaderTenantIdProducer if the header is not set
> 
> What about just setting the tenant to "hawkular" to stay backward compatible?

I think the consesus amongst the components was to return 400 bad request when the header is missing.

metlos:
> I throw an IllegalStateException from the HttpHeaderTenantIdProducer if the header is not set d6ace57#diff-8563bf0980d454e58e40a903d0fd4836R50 That makes the request return 500, which I admit is not optimal. Can you see a way to return a better code with a better message?

You can throw for example `org.hawkular.inventory.api.ValidationException` or even just `IllegalArgumentException`. Those get translated into 400.

Take a look at `org.hawkular.inventory.rest.exception.mappers` package for how different exceptions get translated into responses.

metlos:
> The original Accounts-based InventorySecurity used to autocreate the tenant. d6ace57#diff-e496894d9c6dba4ce79ea79691d9bfbcR126 which I had to comment out because I saw no way to inject an inventory bean to InventorySecurity. Should the tenant auto-creation not be solved elsewhere - e.g. in the REST API, e.g. in a ServletFiltet (that could possible be made inactive using e.g. a System property)?

The filter is not a bad idea. Just make sure you don't do the check for the same tenant on every request, because that check is quite heavy...

pilhuhn:
Tenant creation could be done based on reading a config file and by creating the default 'hawkular' tenant at the initial start.
The list of tenants = short list of strings could be easily cached in memory then.

metlos:
I think metrics auto-create tenants, too... I think all components should have the same semantics...

jkremser:
> Have we ever set "hawkular" as a tenantId header value in the past?

Nope, inventory wasn't using the Hawkular-Tenant header at all (afaik), the tenant was found out from the credentials using the accounts (afaik). And that's what I am talking about. Right now, the tenant header doesn't have to be there in the request. But if we return some 400 code if it's not there in the next version, we are breaking the API. Any old client/curl that's not sending this header will not work anymore. Imho, using some default 'hawkular' tenant is better and if the header is there, then using the tenant from the header.

> I think the consesus amongst the components was to return 400 bad request when the header is missing.

This is imho a stupid consensus :] I can't find it on the devlist. It's just breaking the REST API. Similarly as metrics broke the api when the tenant header went away and they (or accounts) were checking if it was there. If it was there, it was considered as a fail => breaking everything

ppalaga:
> inventory wasn't using the Hawkular-Tenant header at all (afaik), the tenant was found out from the credentials using the accounts (afaik).

Yes, that was true for all Inventory installations in the past, because all of them relied on Accounts. What we do now, is to make the security pluggable, an Accounts-based security being one of the implementations. The Accounts-based security will be used inside Hawkular Community distribution and there will be no API breakage there.

I admit, that with Permissive Security, the API will behave differently, requiring the Hawkular-Tenant header. Is that not acceptable like that?

ppalaga:
4ea5d0a is a squash of all previous commits plus comments accepted plus rebased on top of the latest master. @metlos please review and eventually merge.

metlos:
@ppalaga some build problems in travis.. Maven complains about missing versions...

ppalaga:
ah, I rebased without noticing that you increased minor...

hawkular-inventory-feature-pack/src/main/xsl/subsystem-templates/hawkular-inventory-logging.xsl
```diff
@@ -18,47 +18,31 @@
...
-      </xsl:element>
-    </xsl:copy>
-  </xsl:template>
-
-  <xsl:template
-      match="//*[local-name()='config']/*[local-name()='subsystem']/*[local-name()='console-handler']/*[local-name()='level']">
-    <xsl:copy>
-      <xsl:apply-templates select="@*|node()"/>
-      <xsl:attribute name="name">DEBUG</xsl:attribute>
+      <logger category="org.hawkular.inventory">
```
metlos, line 30: Have you checked the output of this? In the past, I've seen this being translated into the standalone.xml with some superfluous (or even incorrect) xml namespace declarations. Which is why this was originally done the more verbose way using xsl.

hawkular-inventory-feature-pack/src/main/xsl/subsystem-templates/hawkular-inventory-messaging-activemq.xsl
```diff
@@ -18,26 +18,19 @@
...
-       way of saying of xPath's 2.0 prefix-less selector //*:config/*:supplement[@name='default']  -->
-  <xsl:template
-      match="//*[local-name()='config']/*[local-name()='subsystem']/*[local-name()='server' and @name='default']/*[local-name()='jms-topic' and @name='HawkularAccountsEvents']">
+  <xsl:template match="//*[local-name()='config']/*[local-name()='subsystem']/*[local-name()='server'][@name='default']/*[local-name()='jms-queue'][last()]">
     <xsl:copy-of select="."/>
-    <xsl:element name="jms-topic" namespace="{namespace-uri()}">
-      <xsl:attribute name="name">HawkularInventoryChanges</xsl:attribute>
-      <xsl:attribute name="entries">java:/topic/HawkularInventoryChanges</xsl:attribute>
-    </xsl:element>
+    <jms-topic name="HawkularInventoryChanges" entries="java:/topic/HawkularInventoryChanges" />
```
metlos, line 27: Same here - I'm ok with this iff the output in standalone.xml looks "natural".

hawkular-inventory-feature-pack/src/main/xsl/subsystem-templates/hawkular-inventory-logging.xsl
ppalaga, line 30: > Have you checked the output of this? In the past, I've seen this being translated into the standalone.xml with some superfluous (or even incorrect) xml namespace declarations. Which is why this was originally done the more verbose way using xsl.

Yes, the output is correct. `exclude-result-prefixes="xalan"` on L21 does the trick.

hawkular-inventory-api/src/main/java/org/hawkular/inventory/api/Inventory.java
```diff
@@ -103,6 +108,30 @@
...
+                    Tenant.class,
+                    Environment.class,
+                    Feed.class,
+                    ResourceType.class,
+                    MetricType.class,
+                    Resource.class,
+                    Metric.class,
+                    OperationType.class,
+                    MetadataPack.class)
+            .collect(Collectors.toMap(Class::getName, Function.identity()));
```
metlos: This is different from what the `SecurityIntegration#getClassFromName()` did. The original method looked up by lower-cased **simpleName** of the entity types, while you're using the full class name.

hawkular-inventory-feature-pack/src/main/xsl/subsystem-templates/hawkular-inventory-logging.xsl
```diff
@@ -18,47 +18,31 @@
...
+        <level name="${{hawkular.log.inventory:INFO}}" />
+      </logger>
+      <logger category="org.hawkular.inventory.rest.requests">
+        <level name="${{hawkular.log.inventory.rest.requests:INFO}}" />
+      </logger>
+      <logger category="org.hawkular.inventory.rest.interceptors.AutocreateTenantRequestFilter">
+        <level name="TRACE" />
+      </logger>
```
ppalaga, line 38: `TRACE` for `org.hawkular.inventory.rest.interceptors.AutocreateTenantRequestFilter` was not meant to be committed from my side. But actually, it makes no harm in tests. May I leave it like this?

hawkular-inventory-feature-pack/src/main/xsl/subsystem-templates/hawkular-inventory-messaging-activemq.xsl
ppalaga, line 27: Yes, it is natural due to `exclude-result-prefixes="xalan"`

hawkular-inventory-feature-pack/src/main/xsl/subsystem-templates/hawkular-inventory-logging.xsl
metlos, line 38: :+1:

hawkular-inventory-api/src/main/java/org/hawkular/inventory/api/Inventory.java
ppalaga: Good catch. Will post a fix.

pilhuhn:
On 12 May 2016, at 16:47, Jirka Kremser wrote:

> > I think the consesus amongst the components was to return 400 bad 
> > request when the header is missing.
> 
> This is imho a stupid consensus :] I can't find it on the devlist. 
> It's just

It is the "Tenancy on Hawkular Services" thread:
http://lists.jboss.org/pipermail/hawkular-dev/2016-May/002618.html

The reason for the 400 is a scenario where all clients request
(configured) tenant 'bar', but agent is delivering data to 
(non-registered)
tenant 'foo'. The users of 'bar' will never see the data delivered and 
don't
even have a clue that a tenant of 'foo' does not even exist.

PR #263, opened 2016-06-07
HWKINVENT-184 - Removed auth for /status

jpkrohling:
Commons version has to be updated, once hawkular/hawkular-commons#72 gets merged, if the Tenant filter from that PR is desired. Otherwise, the only change required for this is the change on `web.xml`.

### DIFF
--- a/hawkular-inventory-rest-api/pom.xml
+++ b/hawkular-inventory-rest-api/pom.xml
@@ -181,6 +181,12 @@
     </dependency>
 
     <dependency>
+      <groupId>org.hawkular.commons</groupId>
+      <artifactId>hawkular-tenant-jaxrs-filter</artifactId>
+      <version>${version.org.hawkular.commons}</version>
+    </dependency>
+
+    <dependency>
       <groupId>junit</groupId>
       <artifactId>junit</artifactId>
       <scope>test</scope>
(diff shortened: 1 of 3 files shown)


jpkrohling:
@metlos , could you please review?

jpkrohling:
Related to my first comment on this PR: this pulls a new JAX-RS filter from commons, and it will reject all requests to all endpoints/methods if the request has no Hawkular-Tenant. Endpoints/methods can bypass this by annotating the endpoint/method with `@TenantRequired(false)`. 

This effectively means that:
- If there are endpoints/methods currently _not_ requiring `Hawkular-Tenant`, those will be affected
- If all current endpoints/methods are requiring `Hawkular-Tenant`, there's no bad side effect
- For requests without `Hawkular-Tenant`, the filter will reject the request before it reaches the CDI injection phase, so, it will fail fast (and consuming less resources, as the Tenant producer throws an exception).

jpkrohling:
It's now ready for merging after the review.

metlos:
Hmm.. so the dependency on the tenant jaxrs filter was added to automatically check for the presence of the tenant header, by default on all endpoints. 

To not require auth on /status, you simply configure it with no required roles...

~~But where do you say that /status also doesn't require the tenant header?~~
Am I blind or what? :wink:

jpkrohling:
You consume the status handler from commons, and the newest version of commons already has a `@TenantRequired(false)`: https://github.com/hawkular/hawkular-commons/blob/master/hawkular-rest-status/src/main/java/org/hawkular/commons/rest/status/RestStatusHandler.java#L56

pilhuhn:
Yes, explicit allow all is important. The default should be to check for tenant. Otherwise we will introduce holes over time.

metlos:
Agreed.

PR #265, opened 2016-06-14
Allow to make some URI exception from the AutocreateTenantRequestFilt…

jkremser:
…er so that request to /status will not create the tenant. Currently the filter may fail when the tenant header is not there although it should work for unsecured whitelisted endpoints.

cc @jpkrohling

### DIFF
--- a/hawkular-inventory-rest-api/src/main/java/org/hawkular/inventory/rest/interceptors/AutocreateTenantRequestFilter.java
+++ b/hawkular-inventory-rest-api/src/main/java/org/hawkular/inventory/rest/interceptors/AutocreateTenantRequestFilter.java
@@ -17,9 +17,14 @@
 package org.hawkular.inventory.rest.interceptors;
 
 import java.io.IOException;
+import java.util.List;
 import java.util.Set;
 import java.util.concurrent.ConcurrentHashMap;
+import java.util.regex.Pattern;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
 
+import javax.enterprise.inject.Instance;
 import javax.inject.Inject;
 import javax.ws.rs.container.ContainerRequestContext;
 import javax.ws.rs.container.ContainerRequestFilter;
@@ -46,6 +51,10 @@ import org.jboss.resteasy.annotations.interception.ServerInterceptor;
 @ServerInterceptor
 public class AutocreateTenantRequestFilter implements ContainerRequestFilter {
 
+    /* URI chunks to which this filter should not be applied */
+    private static final List<Pattern> uriExceptionPatterns = Stream.of(".*/inventory/status/?",
+            ".*/inventory/ping/?", ".*/inventory/?").map(Pattern::compile).collect(Collectors.toList());
+
     private static final RestApiLogger log =
             Logger.getMessageLogger(RestApiLogger.class, AutocreateTenantRequestFilter.class.getName());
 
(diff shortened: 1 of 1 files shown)


hawkular-inventory-rest-api/src/main/java/org/hawkular/inventory/rest/interceptors/AutocreateTenantRequestFilter.java
```diff
@@ -46,14 +51,27 @@
...
+    /* list of URI chunks to which this filter should not be applied */
+    private static final List<String> uriExceptions = Arrays.asList(".*/inventory/status/?", ".*/inventory/ping/?",
+            ".*/inventory/?");
+    private static List<Pattern> uriExceptionPatterns;
+
+    static {
+        try {
+            uriExceptionPatterns = uriExceptions.stream().map(Pattern::compile).collect(Collectors.toList());
+        } catch (Throwable t) {
+            // swallow, we don't want to fail when classloading
```
abonas: if this fails, and noone knows about it, then won't [this](https://github.com/hawkular/hawkular-inventory/pull/265/files#diff-4b6fb4d559ca774a421a84a5c995b9e5R113) fail later anyway too?

hawkular-inventory-rest-api/src/main/java/org/hawkular/inventory/rest/interceptors/AutocreateTenantRequestFilter.java
```diff
@@ -46,14 +51,27 @@
 @ServerInterceptor
 public class AutocreateTenantRequestFilter implements ContainerRequestFilter {
 
+    /* list of URI chunks to which this filter should not be applied */
+    private static final List<String> uriExceptions = Arrays.asList(".*/inventory/status/?", ".*/inventory/ping/?",
+            ".*/inventory/?");
+    private static List<Pattern> uriExceptionPatterns;
+
+    static {
```
abonas: why is there a need in a static initializer for this code? (and not during constructor for example)

metlos: IMHO this is OK. We don't want to recompile the regexes on every instantiation of this filter (and we don't know how many instances there will be, because that is in the hands of the container). The regex patterns are thread safe objects so having these as a constant is imho preferable to instance field.

hawkular-inventory-rest-api/src/main/java/org/hawkular/inventory/rest/interceptors/AutocreateTenantRequestFilter.java
```diff
@@ -46,14 +51,27 @@
 @ServerInterceptor
 public class AutocreateTenantRequestFilter implements ContainerRequestFilter {
 
+    /* list of URI chunks to which this filter should not be applied */
+    private static final List<String> uriExceptions = Arrays.asList(".*/inventory/status/?", ".*/inventory/ping/?",
+            ".*/inventory/?");
+    private static List<Pattern> uriExceptionPatterns;
```
metlos: should also be `final`. Also, I'd change the def to:

``` java
private static final List<Pattern> uriExceptionPatterns = Stream.of(".*/inventory/status/?", ".*/inventory/ping/?", ".*/inventory/?").map(Pattern::compile).collect(Collectors.toList());
```

to save on the one unnecessary field.

PR #267, opened 2016-06-22
Swagger fixes

metlos:
- Use only definitions from `swagger.json` in the final output, because the stuff generated from the annotations is wrong.
- made the `swagger.json` validate. `editor.swagger.io` should detect no errors in the schema (just warnings about unused definitions, but that's actually intentional, because we really don't mention all the definitions in the URLs - swagger is confused by such an advanced concept as inheritance).
- specified required fields for various entity types
- consolidated POSTs into one endpoint spec for all entity types
- added missing parameter definitions
- added the definition of `InventoryStructure`

### DIFF
--- a/hawkular-inventory-rest-api/pom.xml
+++ b/hawkular-inventory-rest-api/pom.xml
@@ -453,7 +453,13 @@
                       def base = js.parse(bsf)
                       def updates = js.parse(sf)
 
+                      //deep copy
+                      def defs = evaluate(base["definitions"].inspect())
+
                       Updater.updateBase(base, updates)
+
+                      //the definitions in the deprecated API are royally messed up.
+                      base["definitions"] = defs
 
                       def json = groovy.json.JsonOutput.toJson(base)
                       json = groovy.json.JsonOutput.prettyPrint(json)
--- a/hawkular-inventory-rest-api/src/main/rest-doc/swagger.json
+++ b/hawkular-inventory-rest-api/src/main/rest-doc/swagger.json
@@ -73,7 +73,7 @@
         }
       }
     },
-    "/entity/{path:.+}" : {
+    "/entity/{path}" : {
       "get" : {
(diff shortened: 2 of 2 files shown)


metlos:
@pavolloffay would you be able to review, pls?

pavolloffay:
LGTM, Could you please paste somewhere final inventory adoc file?

jkremser:
not sure if it's fixed in the new version, but with the old one I got this:

``` json
{
  "messages": [
    "attribute paths.'/entity/{path:.+}'(get).[path].type is missing",
    "attribute paths.'/entity/{path:.+}'(delete).[path].type is missing"
  ],
  "schemaValidationMessages": [
    {
      "level": "error",
      "domain": "validation",
      "keyword": "oneOf",
      "message": "instance failed to match exactly one schema (matched 0 out of 2)",
      "schema": {
        "loadingURI": "http://swagger.io/v2/schema.json#",
        "pointer": "/definitions/parametersList/items"
      },
      "instance": {
        "pointer": "/paths/~1entity~1{path:.+}/delete/parameters/0"
      }
    }
  ]
}
```

When trying to put it here: http://petstore.swagger.io
- there was `,` on line 1600, that made the JSON invalid

metlos:
The resulting adoc is here http://pastebin.com/7AREU8JU but that is not what I'd like it to look like. Pending https://github.com/hawkular/hawkular-build-tools/pull/25, the output should look like this instead:  http://pastebin.com/n2BzfYQT

metlos:
Ok, NOW it finally [validates](http://online.swagger.io/validator?url=https://raw.githubusercontent.com/metlos/hawkular-inventory/4c9c7f81ffbfd66252681f18730cf50a3148124c/hawkular-inventory-rest-api/src/main/rest-doc/swagger.json)...

pavolloffay:
:+1:

PR #268, opened 2016-06-22
Fix handling of ids with encoded characters while composing a traversal.

metlos:
Handling of IDs with characters needing url encoding was broken. All other filters seem to already be treated correctly.


jkremser:
yesterday's PR ... 1 line
this one ... 2 lines
tomorrow ... ?

I see where it's going again ;)

lgtm

PR #273, opened 2016-07-01
Make sure to not spawn nested/overlapping transactions.

metlos:
- Properly close streams backed by a query
- Always close paged results
- Propagate the current transaction where possible (using .keepTransaction())
- Always commit the transaction previously started (d'oh ;) )
- Don't call methods spawning new transactions from within a transactional
  context
- Added trace logging for transaction boundaries

### DIFF
--- a/hawkular-inventory-api/src/main/java/org/hawkular/inventory/api/model/InventoryStructure.java
+++ b/hawkular-inventory-api/src/main/java/org/hawkular/inventory/api/model/InventoryStructure.java
@@ -22,7 +22,7 @@ import java.util.Collections;
 import java.util.EnumMap;
 import java.util.HashMap;
 import java.util.HashSet;
-import java.util.Iterator;
+import java.util.List;
 import java.util.Map;
 import java.util.Set;
 import java.util.Spliterator;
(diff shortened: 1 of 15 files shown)


hawkular-inventory-api/src/main/java/org/hawkular/inventory/api/model/InventoryStructure.java
```diff
@@ -381,8 +392,23 @@ XBB fillCommon(X entity, XBB bld) {
...
+                //of the txs that we might be using) during its close method. As such, if we just concat'ed our calls
+                //to get children, we might try to nest the transactions - which is not supported by Titan at least.
+                //To overcome this, eagerly evaluate each of our children and compose the resulting stream from the
+                //loaded results.
+                List<?> res;
+                try (Stream<?> next = getChildren(parent, Entity.entityTypeFromSegmentType(st))) {
+                    res = next.collect(Collectors.toList());
+                }
+
+                ret = Stream.concat(ret, res.stream());
```
jkremser, line 411: Is there any impact on performance?

metlos, line 411: The results are in essence iterated over twice and they all reside in memory, so yes.

PR #275, opened 2016-07-21
Closing the txs properly during paging and generic query execution.

metlos:
Pages and query execution would not close transactions if an error occurred.


jkremser:
lgtm

PR #276, opened 2016-07-26
[HAWKULAR-1099] Let the backend decide whether an explicit rollback

metlos:
is required after a commit failure.

### DIFF
--- a/hawkular-inventory-impl-tinkerpop-parent/hawkular-inventory-impl-tinkerpop-titan-provider/src/main/java/org/hawkular/inventory/impl/tinkerpop/provider/TitanProvider.java
+++ b/hawkular-inventory-impl-tinkerpop-parent/hawkular-inventory-impl-tinkerpop-titan-provider/src/main/java/org/hawkular/inventory/impl/tinkerpop/provider/TitanProvider.java
@@ -39,6 +39,7 @@ import com.thinkaurelius.titan.core.Cardinality;
 import com.thinkaurelius.titan.core.Multiplicity;
 import com.thinkaurelius.titan.core.PropertyKey;
 import com.thinkaurelius.titan.core.SchemaViolationException;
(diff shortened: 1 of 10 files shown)


hawkular-inventory-impl-tinkerpop-parent/hawkular-inventory-impl-tinkerpop-titan-provider/src/main/java/org/hawkular/inventory/impl/tinkerpop/provider/TitanProvider.java
```diff
@@ -248,4 +249,17 @@ public RuntimeException translateException(RuntimeException inputException, Cano
...
+
+    @Override public boolean requiresRollbackAfterFailure(Throwable t) {
+        while (t != null) {
+            if (t instanceof TitanException
+                    && "Could not commit transaction due to exception during persistence".equals(t.getMessage())) {
+                return false;
+            }
+
+            t = t.getCause();
+        }
```
jkremser, line 261: FYI https://google.github.io/guava/releases/snapshot/api/docs/com/google/common/base/Throwables.html#getRootCause(java.lang.Throwable)

But I guess, guava is not on the classpath, or is it?

metlos, line 261: it's not on cp..

metlos, line 261: Also that the exception we're checking there is not the root cause unfortunately...

PR #278, opened 2016-08-10
We now produce 3 kinds of hashes for different entities.

metlos:
- All entity types but metdatapack have a content hash - this is just a
  hash of its properties, generic properties and name and is not directly
  used in the API for anything. It is NOT hierarchical, i.e. when a content
  hash changes on a contained entity, the content hash of the containing
  entity doesn't change.
- Identity hash works the same as before and is used for locating the
  "identical" entities - the ones with the same structure, but potentially
  different name or generic properties.
- Sync hash is a new kind of hash that is the union of content and identity
  hashes. It is used as the basis for synchronization of inventory structure.
  I.e. even a changes in generic properties or name of an entity is now
  synchronized.

### DIFF
--- a/hawkular-inventory-api/src/main/java/org/hawkular/inventory/api/model/AbstractHashTree.java
+++ b/hawkular-inventory-api/src/main/java/org/hawkular/inventory/api/model/AbstractHashTree.java
@@ -1,0 +1,283 @@
+/*
+ * Copyright 2015-2016 Red Hat, Inc. and/or its affiliates
+ * and other contributors as indicated by the @author tags.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.hawkular.inventory.api.model;
+
+import java.io.Serializable;
+import java.util.Collection;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.function.BiFunction;
+
+import org.hawkular.inventory.api.TreeTraversal;
+import org.hawkular.inventory.paths.Path;
+import org.hawkular.inventory.paths.RelativePath;
+
+/**
+ * An abstract base class for the various hash trees. This is not instantiable/subclassable by the users on purpose
+ * because only concrete subclasses defined in this package make sense for use. Nevertheless this class is public to
+ * enable abstracting over the different subclasses.
+ *
+ * @author Lukas Krejci
+ * @since 0.18.0
+ */
+public abstract class AbstractHashTree<This extends AbstractHashTree<This, H>, H extends Serializable>
+        implements Serializable {
+    protected final RelativePath path;
+    protected final H hash;
+    protected final Map<Path.Segment, This> children;
+
+    //jackson support
+    AbstractHashTree() {
+        this(null, null, null);
+    }
+
+    AbstractHashTree(RelativePath path, H hash, Map<Path.Segment, This> children) {
+        this.path = path;
+        this.hash = hash;
+        this.children = children;
+    }
+
+    public Collection<This> getChildren() {
+        return children.values();
+    }
+
+    public This getChild(Path.Segment path) {
+        return children.get(path);
+    }
+
+    public H getHash() {
+        return hash;
+    }
+
+    public RelativePath getPath() {
+        return path;
+    }
+
+    public TreeTraversal<This> traversal() {
+        return new TreeTraversal<>(t -> t.children.values().iterator());
+    }
+
+    @Override
+    public boolean equals(Object o) {
+        if (this == o) return true;
+        if (o == null || getClass() != o.getClass()) return false;
+
+        AbstractHashTree<?, ?> tree = (AbstractHashTree<?, ?>) o;
+
+        return hash.equals(tree.hash);
+    }
+
+    @Override
+    public int hashCode() {
+        return hash.hashCode();
+    }
+
+    @Override
+    public String toString() {
+        return "Tree[" + "path=" + path +
+                ", hash='" + hash + '\'' +
+                ",children=" + children.values() +
+                ']';
+    }
+
+    interface TreeConstructor<T extends AbstractHashTree<T, H>, H extends Serializable> {
+        T construct(RelativePath path, H hash, Map<Path.Segment, T> children);
+    }
+
+    /**
+     * An interface that is implemented by all tree hash builders. This is only made public for the cases where the
+     * users would like to abstract over different tree implementations.
+     *
+     * @param <This> the type of the concrete implementation
+     * @param <Child> the type of the tree child builder
+     * @param <Tree> the type of the tree being built
+     * @param <Hash> the type of the hash used in the tree
+     */
+    public interface Builder<This extends Builder<This, Child, Tree, Hash>,
+            Child extends ChildBuilder<Child, This, ?, Tree, Hash>,
+            Tree extends AbstractHashTree<Tree, Hash>, Hash extends Serializable> {
+
+        This withPath(RelativePath path);
+
+        This withHash(Hash hash);
+
+        Hash getHash();
+
+        RelativePath getPath();
+
+        boolean hasChildren();
+
+        Child startChild();
+
+        void addChild(Tree childTree);
+    }
+
+    /**
+     * Interface for top-level hash tree builders. I.e. only these actually have the build method that is used to
+     * produce the final tree.
+     *
+     * @param <This> the type of the concrete implementation
+     * @param <Child> the type of the tree child builder
+     * @param <Tree> the type of the tree being built
+     * @param <Hash> the type of the hash used in the tree
+     */
+    public interface TopBuilder<This extends TopBuilder<This, Child, Tree, Hash>,
+            Child extends ChildBuilder<Child, This, ?, Tree, Hash>,
+            Tree extends AbstractHashTree<Tree, Hash>, Hash extends Serializable>
+            extends Builder<This, Child, Tree, Hash> {
+
+        Tree build();
+    }
+
+    /**
+     * A builder used to build sub trees of some parent tree node.
+     *
+     * @param <This> the type of the concrete implementation of this interface
+     * @param <Parent> the type of the builder of the parent tree node
+     * @param <Child> the type of the builder used to build the subtrees of the tree node built by this builder
+     * @param <Tree> the type of the tree being built
+     * @param <Hash> the type of the hash used in the tree
+     */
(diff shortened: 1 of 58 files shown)


hawkular-inventory-api/src/main/java/org/hawkular/inventory/api/model/AbstractHashTree.java
```diff
@@ -0,0 +1,283 @@
+/*
+ * Copyright 2015-2016 Red Hat, Inc. and/or its affiliates
```
jkremser, line 2: wow, I wonder why the api-checker ok with 2015-2016 for new files?

hawkular-inventory-api/src/main/java/org/hawkular/inventory/api/model/AbstractHashTree.java
```diff
@@ -0,0 +1,283 @@
...
+
+        Parent getParent();
+
+        Parent endChild();
+    }
+
+    static class AbstractBuilder<This extends Builder<This, Child, T, H>,
+            Child extends ChildBuilder<Child, This, ?, T, H>,
+            T extends AbstractHashTree<T, H>, H extends Serializable>
+            implements Builder<This, Child, T, H> {
```
jkremser, line 173: isn't it over-complicated?

hawkular-inventory-api/src/main/java/org/hawkular/inventory/api/model/AbstractHashTree.java
```diff
@@ -0,0 +1,283 @@
...
+
+    abstract static class AbstractChildBuilder<
+            This extends AbstractChildBuilder<This, Parent, Child, T, H>,
+            Parent extends Builder<?, This, T, H>,
+            Child extends AbstractChildBuilder<Child, This, ?, T, H>,
+            T extends AbstractHashTree<T, H>,
+            H extends Serializable>
+
+            extends AbstractBuilder<This, Child, T, H>
+            implements ChildBuilder<This, Parent, Child, T, H> {
```
jkremser, line 263: y41pi4n 1
:D

hawkular-inventory-api/src/main/java/org/hawkular/inventory/api/model/ComputeHash.java
```diff
@@ -0,0 +1,954 @@
...
+
+            @Override
+            public IntermediateHashResult visitEnvironment(Environment.Blueprint environment,
+                                                           IntermediateHashContext context) {
+                return wrap(environment, context, childContext -> {
+                    if (computeContent) {
+                        appendCommonContent(environment, childContext);
+                    }
+                });
+            }
```
jkremser, line 344: Did I miss something, or `computeHash` called on an entity of type Tenant or Environment doesn't call itself for it's ancestors?

hawkular-inventory-api/src/main/java/org/hawkular/inventory/api/model/ComputeHash.java
```diff
@@ -0,0 +1,954 @@
...
+                            .filter(d -> configurationSchema.equals(d.getRole()))) {
+                        return s.findFirst().orElse(dummyDataBlueprint(configurationSchema));
+                    }
+                }
+
+                @Override
+                public DataEntity.Blueprint<?> getConnectionConfiguration(RelativePath root,
+                                                                          Resource.Blueprint parentResource) {
+                    RelativePath resourcePath = root.modified().extend(SegmentType.r, parentResource.getId())
+                            .get().slide(1, 0);
```
jkremser, line 659: these `.slide(1, 0)` cut off the the very first chunk of the canonical path, right?

hawkular-inventory-api/src/main/java/org/hawkular/inventory/api/model/DataEntity.java
```diff
@@ -47,19 +47,22 @@ private DataEntity() {
...
+        super(owner.extend(DataEntity.SEGMENT_TYPE, role.name()).get(), identityHash, contentHash, syncHash, properties);
         this.value = value;
     }
 
-    public DataEntity(CanonicalPath path, StructuredData value, String identityHash, Map<String, Object> properties) {
-        this(path.up(), DataRole.valueOf(path.getSegment().getElementId()), value, identityHash, properties);
+    public DataEntity(CanonicalPath path, StructuredData value, String identityHash, String contentHash,
+                      String syncHash, Map<String, Object> properties) {
+        this(path.up(), DataRole.valueOf(path.getSegment().getElementId()), value, identityHash, contentHash, syncHash,
+                properties);
```
jkremser, line 65: it looks, it's quite common that those `identityHash, contentHash, syncHash` are often used as params together, perhaps introducing some hashBundle object would simplify the method signatures (?)

hawkular-inventory-api/src/main/java/org/hawkular/inventory/api/model/Hashes.java
```diff
@@ -0,0 +1,169 @@
...
+
+/**
+ * A container for all type of hashes computed on an entity. Depending on what kind of hashes have been computed, some
+ * properties of this class might be null. The caller is responsible to keep track of what's been computed and hence
+ * what hashes are held in instances of this class.
+ *
+ * @author Lukas Krejci
+ * @since 0.18.0
+ */
+public final class Hashes implements Serializable {
```
jkremser, line 38: ok, here it is.. forget my previous comment... hmm, well, why this is not used instead of those 3 hash parameters in those methods

hawkular-inventory-api/src/test/java/org/hawkular/inventory/api/test/AbstractBaseInventoryTestsuite.java
```diff
@@ -2554,21 +2569,21 @@ public void testObserveIdentityHashChangedOnParentOfDeletedEntity() throws Excep
...
-                    inventory.observable(Interest.in(Resource.class).having(identityHashChanged())),
-                    inventory.observable(Interest.in(Metric.class).having(identityHashChanged())),
-                    inventory.observable(Interest.in(DataEntity.class).having(identityHashChanged()))
+                    inventory.observable(Interest.in(Feed.class).having(syncHashChanged())),
+                    inventory.observable(Interest.in(ResourceType.class).having(syncHashChanged())),
+                    inventory.observable(Interest.in(MetricType.class).having(syncHashChanged())),
+                    inventory.observable(Interest.in(OperationType.class).having(syncHashChanged())),
+                    inventory.observable(Interest.in(Resource.class).having(syncHashChanged())),
+                    inventory.observable(Interest.in(Metric.class).having(syncHashChanged())),
+                    inventory.observable(Interest.in(DataEntity.class).having(syncHashChanged()))
```
jkremser, line 2586: actually, who can be interested in those hash change events? Is it intentended that agent listen on this and calls the sync if something has changed?

hawkular-inventory-api/src/main/java/org/hawkular/inventory/api/model/SyncHash.java
```diff
@@ -0,0 +1,137 @@
...
+                .treeOf(structure, rootPath, true, true, true, startChild, endChild);
+
+        tbld[0].withPath(res.path).withHash(res.syncHash);
+
+        return ((Tree.Builder)tbld[0]).build();
+    }
+
+    public static final class Tree extends AbstractHashTree<Tree, String> {
+        //jackson support
+        private Tree() {
```
jkremser, line 92: actually, are those 2 new hashes part of the JSON payload if I send a get to the REST api?

jkremser:
Code looks good to me, but I am not aware of the background communication you had with mazz, so perhaps @jmazzitelli could look as well

hawkular-inventory-api/src/main/java/org/hawkular/inventory/api/model/AbstractHashTree.java
metlos, line 2: The license checker actually requires it, I think.

hawkular-inventory-api/src/main/java/org/hawkular/inventory/api/model/AbstractHashTree.java
metlos, line 173: It is. I wanted to be DRY as much as possible, but maybe something like this could also be achieved with overriden methods with specialized return types at the expense of more code.

hawkular-inventory-api/src/main/java/org/hawkular/inventory/api/model/ComputeHash.java
metlos, line 344: You're right and it is intentional :smile: The content hash is not hierarchical. Actually the content hash computation should not call itself for it's descendants (which I think you really meant).

hawkular-inventory-api/src/main/java/org/hawkular/inventory/api/model/ComputeHash.java
metlos, line 659: Yes.

hawkular-inventory-api/src/main/java/org/hawkular/inventory/api/model/DataEntity.java
metlos, line 65: Yeah - I even have that - `Hashes` it's called, but I wanted the constructors to be very explicit. Users don't actually construct entity instances usually - they work with blueprints and updates and obtain the instances of the entities from the API so I consider this more of an implementation detail.

hawkular-inventory-api/src/test/java/org/hawkular/inventory/api/test/AbstractBaseInventoryTestsuite.java
metlos, line 2586: The intention is that anyone - either glue or agent can react on changes. They can react on content hash change if they want to react on the entity itself changing in some way (probably this is identical to reacting to update). They can react on identity hash change if they want to know if the entity itself or any of its children has changed structurally, or they can react on sync hash change if they want to react on the entity or any of its children being changed in any way.

hawkular-inventory-api/src/main/java/org/hawkular/inventory/api/model/AbstractHashTree.java
jkremser, line 263: I forgot there is a AbstractMemeFactoryBuilder on the internets :]
https://memegenerator.net/instance/70760251

hawkular-inventory-api/src/main/java/org/hawkular/inventory/api/model/SyncHash.java
metlos, line 92: Not at the moment. The `/entity/.../treeHash` returns only the `SyncHash.Tree` object that only contains the sync hash, not the other kinds of hashes.

hawkular-inventory-api/src/main/java/org/hawkular/inventory/api/model/ComputeHash.java
jkremser, line 344: right, "descendants" (I english good not)

hawkular-inventory-api/src/main/java/org/hawkular/inventory/api/model/AbstractHashTree.java
metlos, line 263: Well, `class A<This extends A<This>>` is a construct used for "static polymorphism". See for example here: https://en.wikipedia.org/wiki/Curiously_recurring_template_pattern

metlos:
We need to push this forward, so I'll just assume that silence is acceptance :grinning:

PR #279, opened 2016-08-10
Treehash in json

metlos:
Fixes a bug which prevented `/entity/.../treeHash` endpoint from working.

### DIFF
--- a/hawkular-inventory-json-helper/src/test/java/org/hawkular/inventory/json/SerializationTest.java
+++ b/hawkular-inventory-json-helper/src/test/java/org/hawkular/inventory/json/SerializationTest.java
@@ -55,6 +55,7 @@ import org.hawkular.inventory.api.model.DataEntity;
 import org.hawkular.inventory.api.model.Entity;
 import org.hawkular.inventory.api.model.Environment;
 import org.hawkular.inventory.api.model.Feed;
+import org.hawkular.inventory.api.model.IdentityHash;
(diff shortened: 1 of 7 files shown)


hawkular-inventory-json-helper/src/test/java/org/hawkular/inventory/json/SerializationTest.java
```diff
@@ -571,6 +573,46 @@ public void testInventoryStructure() throws Exception {
...
+                .addChild(MetricType.Blueprint.builder(GAUGE).withId("metricType").withUnit(MetricUnit.NONE)
+                        .withInterval(0L).build())
+                .addChild(Metric.Blueprint.builder().withId("metrics").withMetricTypePath("metricType").withInterval
+                        (0L).build())
+                .startChild(
+                        Resource.Blueprint.builder().withId("resource").withResourceTypePath("resourceType").build())
+                .addChild(Resource.Blueprint.builder().withId("childResource").withResourceTypePath("../.resourceType")
+                        .build())
+                .end()
+                .build();
```
jkremser, line 609: hmm, not very DRY

metlos, line 609: Not sure what you mean. If this is duplicated somewhere in the class then I can look to change it, but then again, these are tests, so the locality of test setup and execution plays a role here, too, IMHO.

PR #281, opened 2016-08-15
[HWKINVENT-198] Support restricting synced types.

metlos:


jkremser:
this needs a rebase

metlos:
rebased

hawkular-inventory-api/src/main/java/org/hawkular/inventory/api/model/InventoryStructure.java
```diff
@@ -499,6 +504,11 @@ void impl(Class<E> childType, RelativePath.Extender parent) {
                 blueprints.put(entityPath, entity.entity);
 
                 RelativePath parent = entityPath.up();
+                if (parent.equals(entityPath)) {
+                    //if we can no longer go up, don't add this entity as a child of its parent. It would add itself
+                    //as its own child...
```
jkremser, line 509: adding itself as its own child? What does it mean? a self-loop via contains?

hawkular-inventory-api/src/main/java/org/hawkular/inventory/api/model/SyncConfiguration.java
```diff
@@ -0,0 +1,177 @@
...
+    /**
+     * When syncing entities under some sync root, the user can restrict what types of entities
+     * will be synced. I.e. you can issue a sync of a feed but restrict the sync only to resource types. This will in
+     * effect make sure that the feed only defines the resource types you instruct it to but all the metric types,
+     * resource, metrics, etc. that exist under the feed but were not mentioned in the provided inventory structure
+     * will be left intact.
+     *
+     * @return the set of entity types to be considered during synchronization
+     */
+    public Set<SegmentType> getSyncedTypes() {
```
jkremser, line 61: what does empty set mean? Sync all or sync none

hawkular-inventory-api/src/main/java/org/hawkular/inventory/api/model/SyncConfiguration.java
```diff
@@ -0,0 +1,177 @@
...
+        }
+
+        public Builder withTypes(Iterable<SegmentType> segmentTypes) {
+            segmentTypes.forEach(syncedTypes::add);
+            return this;
+        }
+
+        public Builder withAllTypes() {
+            return withTypes(SegmentType.values());
+        }
```
jkremser, line 152: right, here is the answer

hawkular-inventory-api/src/main/java/org/hawkular/inventory/api/model/InventoryStructure.java
metlos, line 509: The inventory structure object (just that, not the actual DB) would contain such loop, where it would suggest to the callers that the entity indeed contains itself.

hawkular-inventory-api/src/main/java/org/hawkular/inventory/api/model/SyncConfiguration.java
metlos, line 61: sync none, see the `DEFAULT` constant.

PR #282, opened 2016-08-16
[HWKINVENT-190] Make sure to close the query listener JMS connection.

metlos:


jkremser:
I have no idea of the lifecycle of `ConsumerConnectionContext` entity and I wasn't able to find its occurrence in metrics or alerts.

However Alerts from some reason use this weirdness around the `ProducerConnectionContext` 
https://github.com/hawkular/hawkular-alerts/blob/master/hawkular-alerts-actions-bus/src/main/java/org/hawkular/alerts/actions/bus/BusActionPluginSender.java#L56:L57

PR #283, opened 2016-08-17
[HWKINVENT-192] Rename MetricType.type to MetricType.metricDataType

metlos:
This PR tries to retain backwards compatibility for the remote clients by keeping the old property names in the JSON output/input.

Additionally, to make the JSON property names, property filtering and sorting consistent, the `With.Properties` and `With.PropertyValues` filters will use the proper internal names when constructing queries.

### DIFF
--- a/hawkular-inventory-api/src/main/java/org/hawkular/inventory/api/Inventory.java
+++ b/hawkular-inventory-api/src/main/java/org/hawkular/inventory/api/Inventory.java
@@ -556,7 +556,7 @@ public interface Inventory extends AutoCloseable, Tenants.Container<Tenants.Read
             }
 
             @Override public B visitMetricType(MetricType type, Void parameter) {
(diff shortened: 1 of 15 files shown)


jmazzitelli:
LGTM. Merged. I'll get the wildfly agent to use the new API.

abonas:
@pilhuhn does this influence ruby gem/requires a change there?

metlos:
@abonas it should not. The JSON representation of the fields should still include the old properties in the original format. The new fields with the new format are merely added. I plan to remove the old properties in version `0.20.0` while this goes into `0.18.0` (if a `1.0.0` should happen before `0.20.0` is due, I'd remove for `1.0.0`, too, to not litter the API of the first stable release).

This should give you time to accommodate for the change (which is for the better, because it increases consistency in naming in a) inventory and b) between inventory and metrics).

PR #284, opened 2016-08-18
[HWKINVENT-202] Restore MetricType and MetricType.Blueprint Java API.

metlos:
And make sure they still (de)serialize to JSON with the deprecated
properties.


hawkular-inventory-api/src/main/java/org/hawkular/inventory/api/model/MetricType.java
```diff
@@ -221,6 +222,15 @@ public MetricDataType getMetricDataType() {
             return metricDataType;
         }
 
+        /**
+         * @deprecated use {@link #getMetricDataType()}
+         */
+        @Deprecated
+        @JsonSerialize(using = ToStringSerializer.class)
```
jkremser, line 229: `ToStringSerializer` looks good

jkremser:
lgtm

PR #286, opened 2016-08-29
Increase the default number of transaction retries

metlos:
and widen and randomize more the exponential back-off of the retry wait time.

The reason for this PR is that the previous defaults weren't able handle the load the agent itests generate. The delay between retries now increases more rapidly and at the same time the delay is randomized (to a greater extent) from the very retry. Therefore there is an increased chance that the retrying transactions from different threads will not coincide and therefore fail again.

### DIFF
--- a/hawkular-inventory-api/src/main/java/org/hawkular/inventory/base/TraversalContext.java
+++ b/hawkular-inventory-api/src/main/java/org/hawkular/inventory/base/TraversalContext.java
@@ -121,7 +121,7 @@ public final class TraversalContext<BE, E extends AbstractElement<?, ?>> {
     }
 
     private static int getTransactionRetries(Configuration configuration) {
-        String retries = configuration.getProperty(BaseInventory.TRANSACTION_RETRIES, "5");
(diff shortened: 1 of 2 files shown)


jkremser:
Why this can't be handled by Titan?

http://s3.thinkaurelius.com/docs/titan/0.5.4/titan-config-ref.html :
`storage.cassandra.astyanax.retry-policy`
`storage.cassandra.astyanax.retry-backoff-strategy`
`storage.read-time`
`storage.write-time`

Do I recall correctly, that the transaction are not handled by Titan at all, and it's all done in the code? Perhaps the config properties above ^ are for the lower layer of abstraction like connection, I am not sure

metlos:
Yes, the transaction handling is explicit in the code - I.e. we don't let Titan to start and commit transactions automagically, we tell it explicitly when to do it.

As for the retries and such - 2 reasons: 
- I was not aware of these properties ;)
- Using our custom retry code, we can also react on inconsistent states in the DB (i.e. if there is some situation in the backend store that stems from its eventually consistent nature, we can retry, too, even before we try to commit).

Note that the second reason is theoretical in master, but just today I wrote code that takes advantage of that ;)

PR #287, opened 2016-08-30
An attempt at better handling of inconsistent state by retrying the tx's.

metlos:


jmazzitelli:
I'm getting errors in my tests still. I don't see the "That is a bug" exception, but the agent is getting "java.net.SocketTimeoutException: timeout" - never saw that happen before. This failed in different itest suites multiple times. I'm going to set inventory logging categories to DEBUG and see what I get (I get no errors from inventory other than those tx ones at startup: "Could not commit transaction [14] due to exception: com.thinkaurelius.titan.diskstorage.locking.PermanentLockingException: Local lock contention"

For the record, I made sure I'm picking up this new code (0.19.0.Final-SNAPSHOT):

mvn dependency:tree | grep inventory
[INFO] |  +- org.hawkular.inventory:hawkular-inventory-api:jar:0.19.0.Final-SNAPSHOT:test
[INFO] |  +- org.hawkular.inventory:hawkular-inventory-json-helper:jar:0.19.0.Final-SNAPSHOT:test
[INFO] |     +- org.hawkular.inventory:hawkular-inventory-dist:war:0.19.0.Final-SNAPSHOT:test
[INFO] +- org.hawkular.commons:hawkular-inventory-paths:jar:0.7.8.Final:test

jmazzitelli:
I am seeing a series of these (note: in my tests, I have set the sysprop to tell it to retry 15 times - this was an earlier attempt to workaround the bug that I left in)

```
136542 [default task-10] DEBUG org.hawkular.inventory.api  - Commit attempt 1/15 failed. Will wait for 312 ms before retrying. The failure message was: Entity with path: /t;itest-rest-tenant/f;b2590102-df76-4886-8941-55b92758cbfe was associated with more than 1 hash node. That is a bug.
org.hawkular.inventory.base.spi.InconsistentStateException: Entity with path: /t;itest-rest-tenant/f;b2590102-df76-4886-8941-55b92758cbfe was associated with more than 1 hash node. That is a bug.
```

I looked and I see at most 8 retry attempts... "Commit attempt 8/15 failed. Will wait for 133058 ms before retrying." So I guess it thinks it succeeded on the 9th attempt?

jmazzitelli:
So that 8th failed attempt - it says its going to wait 133058 ms?? That's over 2 minutes. Shortly after I see this, in the logs I see my socket timeout. I have a feeling the retry delay might be too long.

metlos:
the socket timeout I assume will be because of the wait time for the request to complete. the transaction retries start at 300ms (+-50%) and each retry is double of the prev value (+-50%). So the 15 retries amount to almost 5 days of max response time... (1st retry waits 300-450ms, the 2nd waits 600-1125ms, 3rd 1200-2813ms, etc.). I spread them so much so that the probability of the retries coinciding in time is lower, but somehow that doesn't seem to help much...

metlos:
My internet connection is super funky right now. The above commit  is my last attempt before the vacation :cry: 

I still get 1 failure in agent's cmdgw itests:

  JdbcDriverCommandITest.testAddJdbcDriver:83 » Throwable Could not get conversa...

but I am not sure if that has anything to do with inventory...

I tried playing with removing unique indices in Titan but that would open up a huge can of worms with inventory needing to deal with way more consistency problems on its own, so I hope that won't be necessary and we find other solution...

metlos:
If this gets past agent itests, would you please merge and release a `0.18.1.Final` please, @Jiri-Kremser ?

jmazzitelli:
I still got similar socket timeout failures with the latest. I'm going to try to run this on my faster laptop and see what happens.

jmazzitelli:
On my faster laptop, I had it pass once, and then fail. Same issues (got the "That is a bug" message during the retries, and eventually agent got socket timeout. I don't think the retries are helping. Whatever the error is, its not recoverable.

```
Commit attempt 11/15 failed. Will wait for 244765 ms before retrying. The failure message was:
Entity with path: /t;itest-rest-tenant/f;7d732eba-9f47-4cb1-b241-458ff5b285b4 was associated
with more than 1 hash node. That is a bug.
```

later followed by the agent error

```
java.net.SocketTimeoutException: timeout
   at okio.Okio$3.newTimeoutException(Okio.java:207)
   at okio.AsyncTimeout.exit(AsyncTimeout.java:261)
   at okio.AsyncTimeout$2.read(AsyncTimeout.java:215)
```

metlos:
May this be remembered as the issue that forced Titan out of Inventory :wink:

PR #288, opened 2016-09-12
Quick port to Tinkerpop3

metlos:
This leaves everything as it was in inventory and just updates the query translation to support Tinkerpop3.

This in turn requires update of Titan to 1.0.0 (which is the latest and most probably last released version).


metlos:
Ok, this does will not pass. I will have to create a single PR with the whole lot to move to sqlg...

PR #291, opened 2016-09-15
[HWKINVENT-8] Versioning

metlos:
Work in progress on versioning.

Main principles:
- Nothing gets deleted, merely is marked as such (a new `eradicate()` method can truly delete stuff, but the backwards compatible `delete()` methods merely mark the elements as deleted from certain point in time onwards). 
- Entities are stored in 2 vertices - an identity vertex containing identifying info on the entity (its canonical path, type, etc) and state vertex that contains the rest of the entity properties as they exist at some point in time.
- The identity and state vertex are connected using an `__inState` edge that has `from` and `to` properties containing time stamps. Those define from and until when is given state valid.
- The current state has `to` set to `Long.MAX_VALUE` - a distant future.
- An entity is non-existent at some point in time if there is no identity vertex at all or there is no valid `__inState` edge for given point in time.
- All relationships have `from` and `to` properties, too, that define from and until when those relationships "exist".
- Relationship edges connect the identity vertices.
- All querying has been updated to take into account the `from` and `to` props.
- The "current" time is either left undefined, in which case we use `Instant.now()`, or is defined using `Inventory.at()` method.
- The current time is passed down to the backend as a `Discriminator` - the reasoning being that in future the discriminator might hold more than just time.

Still to do:
- [x] unit tests for `eradicate()` and "looking into the past"
- [x] /history REST
- [x] `?at=<time>` query param
- [x] create indices for time-related properties and update the pre-created schema with them
- [x] integration tests

### DIFF
--- a/hawkular-inventory-api/src/main/java/org/hawkular/inventory/api/model/Change.java
+++ b/hawkular-inventory-api/src/main/java/org/hawkular/inventory/api/model/Change.java
@@ -24,9 +24,10 @@ import org.hawkular.inventory.api.Action;
  * Represents a change made to an entity at some point in time.
  *
  * @author Lukas Krejci
- * @since 0.19.0
+ * @since 0.20.0
  */
-public final class Change<Element extends AbstractElement<?, ?>, Context> implements Comparable<Change<?, ?>> {
+public final class Change<Element extends Entity<?, ?>, Context>
+        implements Comparable<Change<?, ?>> {
     private final Instant time;
     private final Action<Context, Element> action;
     private final Context actionContext;
@@ -66,7 +67,7 @@ public final class Change<Element extends AbstractElement<?, ?>, Context> implem
             return diff;
         }
 
-        return element.getId().compareTo(o.element.getId());
+        return element.getPath().toString().compareTo(o.element.getPath().toString());
     }
 
     @Override public boolean equals(Object o) {
--- a/hawkular-inventory-api/src/main/java/org/hawkular/inventory/api/model/Change.java
+++ b/hawkular-inventory-api/src/main/java/org/hawkular/inventory/api/model/Change.java
@@ -13,9 +13,7 @@
  * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
  * See the License for the specific language governing permissions and
  * limitations under the License.
- *
  */
-
 package org.hawkular.inventory.api.model;
 
 import java.time.Instant;
--- a/hawkular-inventory-api/src/main/java/org/hawkular/inventory/api/model/Change.java
+++ b/hawkular-inventory-api/src/main/java/org/hawkular/inventory/api/model/Change.java
@@ -1,0 +1,90 @@
+/*
+ * Copyright 2015-2016 Red Hat, Inc. and/or its affiliates
+ * and other contributors as indicated by the @author tags.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ */
+
+package org.hawkular.inventory.api.model;
+
+import java.time.Instant;
+
+import org.hawkular.inventory.api.Action;
+
+/**
+ * Represents a change made to an entity at some point in time.
+ *
+ * @author Lukas Krejci
+ * @since 0.19.0
+ */
+public final class Change<Element extends AbstractElement<?, ?>, Context> implements Comparable<Change<?, ?>> {
+    private final Instant time;
+    private final Action<Context, Element> action;
+    private final Context actionContext;
+    private final Element element;
+
+    public Change(Instant time, Action<Context, Element> action, Context actionContext, Element element) {
+        this.time = time;
+        this.action = action;
+        this.actionContext = actionContext;
+        this.element = element;
+    }
+
+    public Instant getTime() {
+        return time;
+    }
+
+    public Action<Context, Element> getAction() {
+        return action;
+    }
+
+    public Context getActionContext() {
+        return actionContext;
+    }
+
+    public Element getElement() {
+        return element;
+    }
+
+    @Override public int compareTo(Change<?, ?> o) {
+        int diff = time.compareTo(o.time);
+        if (diff != 0) {
+            return diff;
+        }
+
+        diff = action.asEnum().ordinal() - o.action.asEnum().ordinal();
+        if (diff != 0) {
+            return diff;
+        }
+
+        return element.getId().compareTo(o.element.getId());
+    }
+
+    @Override public boolean equals(Object o) {
+        if (this == o) return true;
+        if (o == null || getClass() != o.getClass()) return false;
+
+        Change<?, ?> change = (Change<?, ?>) o;
+
+        return time.equals(change.time) && action.equals(change.action) && element.equals(change.element);
+
+    }
+
+    @Override public int hashCode() {
+        int result = time.hashCode();
+        result = 31 * result + action.hashCode();
+        result = 31 * result + element.hashCode();
+        return result;
+    }
+}
--- a/hawkular-inventory-api/src/main/java/org/hawkular/inventory/api/model/Change.java
+++ b/hawkular-inventory-api/src/main/java/org/hawkular/inventory/api/model/Change.java
@@ -17,6 +17,7 @@
 package org.hawkular.inventory.api.model;
 
 import java.time.Instant;
+import java.util.Objects;
 
 import org.hawkular.inventory.api.Action;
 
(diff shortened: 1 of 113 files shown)


metlos:
@jotak or @Jiri-Kremser could you please start reviewing this? This is a big PR because it's a big new feature :wink: Hopefully looking at it commit by commit will be manageable...

I still need to add integration tests for the new REST endpoints and query params but I'm hopeful that's gonna be relatively smooth.

I'll be offline tomorrow but will be available again on Thu and Fri..

jotak:
Big piece of work! :-) I haven't looked at the code yet, just reacting to the PR description (and thank you for that detailed description)

So if I understand, every modification of the vertex state, or of the relationships, will basically "close" the "to" time of the previous element and create a new element with opened "to". Which means the overall stored data will grow indefinitely, or is there any kind of TTL? Are we fine with that?
Since every query now uses "from" and "to", is the performance impact negligible?

hawkular-inventory-api/src/main/java/org/hawkular/inventory/api/model/Change.java
```diff
@@ -0,0 +1,90 @@
...
+    public Context getActionContext() {
+        return actionContext;
+    }
+
+    public Element getElement() {
+        return element;
+    }
+
+    @Override public int compareTo(Change<?, ?> o) {
+        int diff = time.compareTo(o.time);
```
jotak, line 66: Quick question here... which is more or less related to the code guidelines:
Since there's no null-check on "o": do you globally assume there is never ever "null" objects? In which case I'm fine with no null-check, we just have to be careful on code-review.
Or if there's a mix of sometimes nullable objects, sometimes non-nullable, then it might be appropriate to use checks like guava's "checkNotNull"...

hawkular-inventory-api/src/main/java/org/hawkular/inventory/base/BaseData.java
```diff
@@ -193,34 +191,36 @@ void preCreate(DataModificationChecks<BE> checks, DataEntity.Blueprint<R> bluepr
...
-                Log.LOGGER.wNoDataAssociatedWithEntity(tx.extractCanonicalPath(entityRepresentation));
-                return;
-            }
-
-            BE dataRel = rels.iterator().next();
-
-            BE structuredData = tx.getRelationshipTarget(dataRel);
-
-            tx.deleteStructuredData(structuredData);
+// This is pre-versioning code - all this is only invoked during eradicate now...
```
jotak: It should probably be deleted before merging the PR, no?

jotak (approved):

hawkular-inventory-impl-tinkerpop-parent/hawkular-inventory-impl-tinkerpop/src/main/java/org/hawkular/inventory/impl/tinkerpop/FilterVisitor.java
```diff
@@ -85,7 +90,7 @@ public void visit(GraphTraversal<?, ?> query, Related related, QueryTranslationS
                 if (null != related.getRelationshipName()) {
                     state.setInEdges(true);
                     state.setComingFrom(Direction.IN);
-                    query.inE(related.getRelationshipName());
+                    query.inE(related.getRelationshipName()).discriminate(discriminator);
```
jotak: Isn't it possible to "discriminate" the query globally once for all (and then, the discriminator would need to be propagated in the subsequent traversal). It would be safer than discriminating each case one by one, no?
Of course it would mean that the discriminator cannot be changed during a traversal, but if I understand the concept it's a kind of global setting, no?

hawkular-inventory-api/src/main/java/org/hawkular/inventory/base/BasePreCommit.java
```diff
@@ -534,7 +537,7 @@ boolean canBeIdentityRoot() {
         void loadFrom(Transaction<BE> tx) throws ElementNotFoundException {
             if (element == null || loadingTx != tx) {
                 loadingTx = tx;
-                Discriminator disc = Discriminator.time(Instant.now());
+                Discriminator disc = Discriminator.time(/*isDelete ? txStart :*/ Instant.now());
```
jotak: Can you elaborate, why in case of deletion we may have to use txStart, or not?

jotak (approved):
I assume it's just reformatting

hawkular-inventory-itest/src/test/java/org/hawkular/inventory/rest/test/InventoryITest.java
```diff
@@ -1482,6 +1482,31 @@ public void testTreeHash() throws Throwable {
         assertEquals(tree.getChild(Path.Segment.from("r;table")).getHash(), table.getSyncHash());
     }
 
+    @Test
+    public void testHistory_onEntity() throws Exception {
+        //TODO implement
```
jotak: Don't forget these :)

jotak:
(lol... forget the above comment "I assume it's just reformatting" ; it was about a particular commit, not the whole PR! :smile:)

metlos:
> So if I understand, every modification of the vertex state, or of the relationships, will basically "close" the "to" time of the previous element and create a new element with opened "to". Which means the overall stored data will grow indefinitely, or is there any kind of TTL? Are we fine with that?

There is currently no TTL and data grows indefinitely, yes. I assumed that for a v1 of the functionality this would be sufficient. But I agree that we should put a mechanism in place to delete or compact old entries. Would you mind creating a JIRA for that?

> Since every query now uses "from" and "to", is the performance impact negligible?

Some kind of these checks is going to always be necessary unless we want the whole graph be copy-on-write, which is I think even less practical.

This actually reminded me of one more thing that needs to be done - I forgot to add indices on these properties and update the schema precreation... Adding a bullet point for it in the description.

hawkular-inventory-api/src/main/java/org/hawkular/inventory/api/model/Change.java
metlos: The contract of `compareTo()` is to throw an NPE if the parameter is null, if I am not mistaken. I guess I should add a null checks for the properties of the change to be non-null, though.

hawkular-inventory-api/src/main/java/org/hawkular/inventory/base/BaseData.java
metlos: Yes, forgot to do that...

hawkular-inventory-impl-tinkerpop-parent/hawkular-inventory-impl-tinkerpop/src/main/java/org/hawkular/inventory/impl/tinkerpop/FilterVisitor.java
metlos: This is a leftover from a wrong logic I put in place at some point during the impl. I should revert that...

hawkular-inventory-api/src/main/java/org/hawkular/inventory/base/BasePreCommit.java
metlos: This is a leftover from a wrong logic I put in place at some point during the impl. I should revert that...

hawkular-inventory-impl-tinkerpop-parent/hawkular-inventory-impl-tinkerpop/src/main/java/org/hawkular/inventory/impl/tinkerpop/FilterVisitor.java
metlos: I think it could be done using some decorator strategy of the graph. I didn't actually know about that possibility until a couple of days ago (I thought that Tinkerpop uses graph strategies solely for query optimization but apparently it can also use them for adding "application logic" to the traversals).

That said I am not sure if such global thing would work, because I have a feeling we do a little bit of special handling here and there that does not follow otherwise uniform application of those checks. But it is a great point and we should definitely investigate the possibility.

hawkular-inventory-api/src/main/java/org/hawkular/inventory/api/model/Change.java
jotak: Yes, I put my comment here because that's where I noticed potential NPE but the null-checks, if we decide to do them, should definitely be done in constructors or setters.

It's not a remark for this PR in particular, but more a general thought.

hawkular-inventory-impl-tinkerpop-parent/hawkular-inventory-impl-tinkerpop/src/main/java/org/hawkular/inventory/impl/tinkerpop/FilterVisitor.java
jotak: Well I didn't mean to do something really global, but just in the scope of the "query" object. What I don't know is, if you write "query.discriminate(disc).inE(related.getRelationshipName())" => is the discriminator propagated? In other words, is it equivalent to "query.inE(related.getRelationshipName()).discriminate(disc)" ?

metlos: no, in here `discriminate()` is very similar to `has()`. It only applies to the current position in the traversal.

jotak:
LGTM (but a small one considering the amount of changes! :wink: )

PR #289, opened 2016-09-13
Move to sqlg

metlos:
This is quite large, because it does a couple of things unfortunately.
- The `TinkerpopBackend` has been ported to Tinkerpop3.
- I tried move Titan forward, too, but the poor thing is so unmaintained, I need it built from its unmerged PR (with very little chance of it getting merged). So in the end, Titan is disabled altogether.
- Change the build so that we define the datasources and include the jdbc drivers of the supported databases.

I've tried splitting this PR into smaller pieces but they would just not pass the tests. Therefore I humbly ask for this to be reviewed commit by commit.

PR #292, opened 2016-09-16
Move to sqlg

metlos:
## NOTE: This is a replacement for #289 being merged from a branch in the main repo so that more people can easily collaborate...

This is quite large, because it does a couple of things unfortunately.
- The `TinkerpopBackend` has been ported to Tinkerpop3.
- I tried move Titan forward, too, but the poor thing is so unmaintained, I need it built from its unmerged PR (with very little chance of it getting merged). So in the end, Titan is disabled altogether.
- Change the build so that we define the datasources and include the jdbc drivers of the supported databases.

I've tried splitting this PR into smaller pieces but they would just not pass the tests. Therefore I humbly ask for this to be reviewed commit by commit.


jotak:
LGTM

PR #293, opened 2016-09-16
Fix uniqueness condition in sqlg triggering IllegalArgumentException

jotak:
It must throw an EntityAlreadyExistsException instead

Note that it changes the http code returned in such case (500 vs 409), fixing some itest that was failing

### DIFF
--- a/hawkular-inventory-impl-tinkerpop-parent/hawkular-inventory-impl-tinkerpop-sql-provider/src/main/java/org/hawkular/inventory/impl/tinkerpop/sql/SqlGraphProvider.java
+++ b/hawkular-inventory-impl-tinkerpop-parent/hawkular-inventory-impl-tinkerpop-sql-provider/src/main/java/org/hawkular/inventory/impl/tinkerpop/sql/SqlGraphProvider.java
@@ -32,10 +32,12 @@ import org.apache.commons.configuration.MapConfiguration;
 import org.apache.tinkerpop.gremlin.structure.Graph;
 import org.apache.tinkerpop.gremlin.structure.Vertex;
 import org.hawkular.inventory.api.Configuration;
+import org.hawkular.inventory.api.EntityAlreadyExistsException;
 import org.hawkular.inventory.api.model.Entity;
 import org.hawkular.inventory.impl.tinkerpop.spi.Constants;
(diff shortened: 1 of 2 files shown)


hawkular-inventory-impl-tinkerpop-parent/hawkular-inventory-impl-tinkerpop-sql-provider/src/main/java/org/hawkular/inventory/impl/tinkerpop/sql/SqlGraphProvider.java
```diff
@@ -144,4 +146,16 @@ private static Object sampleValue(Class<?> type) {
             throw new IllegalArgumentException("Unhandled type of property: " + type);
         }
     }
+
+    @Override public RuntimeException translateException(RuntimeException inputException, CanonicalPath affectedPath) {
+        // Sqlg sends IllegalArgumentException when unique constraint is violated
+        // See SqlgElement.updateRow
+        if (inputException instanceof IllegalArgumentException
```
pilhuhn, line 153: Is there no way of throwing that EAEE earlier so that this 'complicated' logic is not needed?

pilhuhn:
@jotak What iTests? Can('t) they be fixed with this commit?

hawkular-inventory-impl-tinkerpop-parent/hawkular-inventory-impl-tinkerpop-sql-provider/src/main/java/org/hawkular/inventory/impl/tinkerpop/sql/SqlGraphProvider.java
metlos: This is coming from the backend (sqlg) right after we try the insert. A different exception is coming from Titan. We have a common logic but ask the providers for backend specific "questions" (like the "what does this exception mean?")..

This is of course brittle and would be best solved by making Sqlg throwing some more specific exception, but then again - Sqlg would have to do the same "magic" because it would have to translate the SQLExceptions thrown from the underlying database - those differ per db vendor, too.

jotak:
Sorry my bad, my description was not clear. My patch fixes the itest, not the contrary :)

PR #294, opened 2016-09-19
Use wildfly-server-provisioning-maven-plugin

jotak:
Create a full distribution under hawkular-inventory-feature-pack when profile "server-provisioning" is active.

So that inventory can run out of the box, without Services, with all wildfly dependencies provided.

### DIFF
--- a/hawkular-inventory-feature-pack/pom.xml
+++ b/hawkular-inventory-feature-pack/pom.xml
@@ -251,4 +251,30 @@
 
     </plugins>
   </build>
+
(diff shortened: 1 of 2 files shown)


metlos:
Sorry I didn't comment in time on IRC - I'd rather see this in `hawkular-inventory-dist` in a standalone profile.

jotak:
actually this is useless... I ended up doing the same thing that hawkular-inventory-itest already does.

PR #295, opened 2016-09-20
Pre-create the Inventory schema in SQL so that it is not dynamically

metlos:
created during the heavy load the feeds can inflict on the DB during sync.

Also, use my fork of Sqlg so that the methods that we use for schema
creation are public (this will eventually go into Sqlg proper at which
point we can switch back).


metlos:
@jotak or @Jiri-Kremser, could you please take a look at this? Once this is in, I want to release `0.19.2.Final` so that we can finally release services without Titan (barring additional undiscovered problems, of course ;) ).

jotak:
Looks good to me

PR #296, opened 2016-09-21
[HWKINVENT-187] Remove unnecessary deps gremlin-groovy

jotak:
This dependency isn't used in sqlg for our purpose and brings a lot of groovy jars
WAR size decreased by 12MB (35 to 23)

(Note: I will suggest some change directly in sqlg to avoid doing exclusion here)


metlos:
:+1:

PR #297, opened 2016-09-28
[HWKINVENT-204] Send Bad Request on treeHash / environment

jotak:
Quick fix to send more explicit error (before that, user got a ClassCastException stack in his response)

@metlos I check here that the element type is not "environment", but is there other types that should be disallowed here?


hawkular-inventory-rest-api/src/main/java/org/hawkular/inventory/rest/RestEntity.java
```diff
@@ -71,6 +72,9 @@ public RestEntity() {
         CanonicalPath path = CanonicalPath.fromPartiallyUntypedString(getPath(uriInfo, "/treeHash".length()),
                 getTenantPath(), AbstractElement.class);
 
+        if (path.getSegment().getElementType() == SegmentType.e) {
```
metlos: Another non-synced type is the tenant.

A generic test for this, would, I guess, look like:

``` java
Syncable.class.isAssignableFrom(Inventory.types().byPath(path).getElementType())
```

jotak: done

PR #298, opened 2016-10-04
Make the connection url, user and password configurable ...

metlos:
... also through sysprops and env vars.


pilhuhn:
LGTM
Does it still default to hsqldb if no properties are passed?

metlos:
Yes. If nothing is passed in through command line or env vars, the props are taken from a default configuration file that points to the HSQLDB data source.

metlos:
I also updated the docs on the site to take this into account - https://github.com/hawkular/hawkular.github.io/pull/230

PR #299, opened 2016-10-05
Use the latest Sqlg to reduce the size of the inventory dist.

metlos:


hawkular-inventory-dist/pom.xml
```diff
@@ -138,47 +138,20 @@
...
-          <groupId>org.apache.tinkerpop</groupId>
-          <artifactId>gremlin-groovy</artifactId>
-        </exclusion>
-      </exclusions>
     </dependency>
 
     <dependency>
       <groupId>org.umlg</groupId>
-      <artifactId>sqlg-h2</artifactId>
+      <artifactId>sqlg-h2-dialect</artifactId>
```
pilhuhn: Should we completely remove artifacts for the in-memory store we are not using (i.e. if we use hsqldb, then remove h2-support and vice versa) ?

metlos: Yeah, we probably should. I chose HSQLDB as the default simply because it's more stable in Sqlg but on the other hand, H2 is included in Wfly by default so we would not have to ship the drivers for it. I need to do more testing with H2 backend to see if we could switch to it.

PR #300, opened 2016-10-13
Synchronize the versions of the driver used by the JNDI datasources and the Sqlg managed ones.

metlos:
Note that this PR is against the "stabilization branch" - 1.0.x that is based on 0.19.x - i.e. it does NOT include the versioning support.

This is because we don't need versioning for anything in hawkular yet and will not start using it until the components are stabilized.

### DIFF
--- a/hawkular-inventory-dist/pom.xml
+++ b/hawkular-inventory-dist/pom.xml
@@ -140,12 +140,24 @@
       <groupId>org.umlg</groupId>
       <artifactId>sqlg-hsqldb-dialect</artifactId>
       <version>${version.org.umlg}</version>
+      <exclusions>
+        <exclusion>
+          <groupId>org.hsqldb</groupId>
+          <artifactId>hsqldb</artifactId>
+        </exclusion>
+      </exclusions>
     </dependency>
 
     <dependency>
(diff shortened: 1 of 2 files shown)


hawkular-inventory-dist/pom.xml
```diff
@@ -157,6 +169,19 @@
       <!--<version>${version.org.umlg}</version>-->
     <!--</dependency>-->
 
+    <!-- include the versions of the drivers we want -->
+    <dependency>
+      <groupId>org.postgresql</groupId>
+      <artifactId>postgresql</artifactId>
+      <version>${version.org.postgresql}</version>
```
pilhuhn, line 176: Could that be made 'provided' given that it is loaded as module already?

metlos: There are 2 options here - either have the driver in WEB-INF/lib (and make sure at build time it is the same version as the module we deploy) - that's how it is done now, or we could depend on the driver module and modify the jboss-deployment-structure to reflect that. The second option is more complicated to do in -dist, because we'd need some xml transformation to do add that into the deployment-structure that the rest.war declares. Not impossible, but this was just simpler (at the cost of slightly bigger distribution size).

PR #301, opened 2016-10-17
[HWKINVENT-201] Remove the deprecated REST API.

metlos:
This should have gone into 1.0.0.Final but, well, it didn't :disappointed:


metlos:
Whoops, overwrote this with another PR. Lesson learned for really using the topic branches EVERY TIME ;)

PR #303, opened 2016-10-20
Perf improvements

metlos:
Each commit represents 1 improvement. Review them individually, please.


hawkular-inventory-api/src/main/java/org/hawkular/inventory/base/BasePreCommit.java
```diff
@@ -552,6 +619,30 @@ void add(EntityAndPendingNotifications<BE, ?> entity) {
...
+                for (ProcessingTree<BE> c : children) {
+                    if (expectedCp.equals(c.cp)) {
+                        current = c;
+                        children = c.children;
+                        break;
+                    }
+                }
+            }
+
+            return segments.hasNext() ? null : current;
```
jotak, line 648: Maybe I'm too tired, but... segments.hasNext() will always be false, no?
I'm not sure to understand the algorithm globally... shouldn't we have a breaking condition on the "while" loop?

metlos: Good catch! it now only works by accident because usually ids on different levels are different.

The algorithm tries to go down the provided path and for each segment, it tries to find the appropriate child. If it finds the child, it compares the next segment with its children, etc.

So the missing break condition should be on current being null after the for-loop. As it is right now, it always returns the "current".

jotak:
Ok, looks good to me

PR #308, opened 2016-10-26
H2 as default

metlos:
This seems to be stable memory-wise (to the extent of hours with 10 feeds registering simultaneously shortly after server comes up).


metlos:
@jotak this is ready for review now, too.

PR #309, opened 2016-11-01
Sqlg bump + optimizations

metlos:
Sqlg internals have some range filtering improvements and other general perf improvements.

There's also a gremlin traversal generation optimization in inventory.

Overall we should see better performance with this update.

### DIFF
--- a/.travis.filtered.build.sh
+++ b/.travis.filtered.build.sh
@@ -37,6 +37,9 @@ set -o pipefail
 # /^.* <<< FAILURE! *$/p
 # This how surefire reports errors, so we output that to have some visibility of the failing tests in the travis log.
 #
+# /.* <<< ERROR! */,/^\s*$/p
+# If there is a test error print it and everything under it until the next empty line - this way we see the failed tests.
+#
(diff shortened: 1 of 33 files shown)


jotak:
You had to rollback a sqlg bump previously, is it right? (the thing about light artifacts). The issue you had is now fixed?

metlos:
Hmm... I don't know why Travis is failing this... All is passing locally.
But yes, the issues we had with the previous update seem to have been resolved. But I'm marking this as "don't merge yet" to investigate the failure...

metlos:
Heh. So things worked for me locally, because the code before the latest commit in this PR depended on the wrong sqlg artifact that brought in a snapshot version of the dialect that I built locally and that worked.

Reintroducing the "slimming" commit fixed inv to depend on the correct artifacts and so this passes on Travis now, too.

metlos:
Hmm.. so I see the errors about the cached plan again when connected to postgres.. Need to investigate this...

metlos:
@jotak could you please take another look at this?

jotak:
LGTM

PR #310, opened 2016-11-02
Upgrade rxjava 1.0.8 -> 1.1.7

pgier:
This syncs with metrics and the version can optionally be inherited from the next hawkular parent update.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -58,7 +58,7 @@
   <properties>
     <version.org.hawkular.commons>0.7.4.Final</version.org.hawkular.commons>
 
-    <version.io.reactivex>1.0.8</version.io.reactivex>
+    <version.io.reactivex>1.1.7</version.io.reactivex>
     <!-- overrides 3.1.2 from parent which produces NPE -->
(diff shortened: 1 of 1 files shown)


jkremser:
If we want to assure the same versions of RxJava for all components, shouldn't this go into https://github.com/hawkular/hawkular-parent-pom ?

metlos:
This should go into `1.0.x` branch instead of master. In the meantime I've made `1.0.x` the default branch so the new PRs should be automatically made against that by default. @pgier could you please make a new PR against `1.0.x` and close this PR once that's done?

pgier:
@Jiri-Kremser It's already been added to hawkular-parent-pom, but I wanted to submit the PR here also to make sure there aren't any breaking changes before upgrading to the new parent.

pgier:
PR #311 created against 1.0.x branch.  @metlos wouldn't you want to also upgrade the version in master if we are upgrading in the 1.0.x branch?

metlos:
@pgier, yes, but I want to do it by merging `1.0.x` into master.. Closing this PR and moving on to #311. Thank you!

PR #313, opened 2016-11-09
Performance improvements

metlos:
The query optimizations yield quite impressive results. There are still things to do but I'm creating the PR now so that the progress so far can be reviewed.


jotak (approved):
LGTM

jotak:
thanks @metlos

PR #315, opened 2016-11-16
Swagger: rename MetricType.type to MetricType.metricDataType

jotak:


metlos:
This PR should be done against the `1.0.x` branch which is now the default we develop against.

PR #316, opened 2016-11-18
HWKINVENT-200 Remove the deprecated properties from MetricType (breaking change)

jotak:
MetricType.type is replaced with MetricType.metricDataType. It impacts the REST API.
The deserialization of metricDataType now expects upper case (before, "type" expected upper case and "metricDataType" expected lower case)

Labeled "don't merge yet" to first make sure this is what we want, and also to first make sure clients will be properly updated


jotak:
@metlos can you have a look?

I suggest we use upper case "metricDataType" ; which means we can get rid of the specific mixins written for that enum

hawkular-inventory-json-helper/src/main/java/org/hawkular/inventory/json/InventoryJacksonConfig.java
```diff
@@ -126,7 +123,6 @@ public static void configure(ObjectMapper objectMapper) {
         objectMapper.addMixIn(Feed.class, FeedMixin.class);
         objectMapper.addMixIn(Metric.class, MetricMixin.class);
         objectMapper.addMixIn(MetricType.class, MetricTypeMixin.class);
-        objectMapper.addMixIn(MetricDataType.class, MetricDataTypeMixin.class);
```
jotak, line 129: Actually I can probably delete MetricDataTypeMixin.java

metlos:
> @metlos can you have a look?
> 
> I suggest we use upper case "metricDataType" ; which means we can get rid of the specific mixins written for that enum

If we were to do that, we also need to change how that value is persisted to the database, so that filtering on it works as expected. I.e. the values that appear in JSON should be the same you use
in `/traversal`s. This will only work if that value is stored in database as such. Currently in we store the metric data type using its display name (https://github.com/hawkular/hawkular-inventory/blob/1.0.x/hawkular-inventory-impl-tinkerpop-parent/hawkular-inventory-impl-tinkerpop/src/main/java/org/hawkular/inventory/impl/tinkerpop/TinkerpopBackend.java#L777 and https://github.com/hawkular/hawkular-inventory/blob/1.0.x/hawkular-inventory-impl-tinkerpop-parent/hawkular-inventory-impl-tinkerpop/src/main/java/org/hawkular/inventory/impl/tinkerpop/TinkerpopBackend.java#L978).

jotak:
Ok so I reverted to lower case ; however we may not want to merge it quickly, because it would imply increasing version due to API breaking change

jotak:
@metlos, the last commit adds a "toString" to enum MetricDataType. I haven't find explicit or implicit usages of "toString", except in unit tests ContentHashTest.java and IdentityHashTest.java. You think it's ok?

metlos:
> @metlos, the last commit adds a "toString" to enum MetricDataType. I haven't find explicit or implicit usages of "toString", except in unit tests ContentHashTest.java and IdentityHashTest.java. You think it's ok?

👍

jotak:
a little bit outdated.....

PR #320, opened 2016-11-30
Implement transitive closure for Cassandra

jotak:


hawkular-inventory-impl-cassandra/src/main/java/org/hawkular/inventory/impl/cassandra/CassandraBackend.java
```diff
@@ -159,8 +170,59 @@ public CassandraBackend(RxSession session) {
...
-        //TODO implement
-        return null;
+        String startingCP = startingPoint.getString(Statements.CP);
+        if (startingCP == null) {
+            // Not a vertex
+            return Collections.<Row>emptyList().iterator();
+        }
+        Observable<Row> closureContent = getRelationshipTables(direction)
+                .flatMap(table -> {
+                    Statement q = select().all().from(table).where(in(Statements.NAME, Arrays.asList(relationshipNames)));
```
jotak: For this algorithm I have to query on "name", which means we should create an index for that (or change the algorithm at the price of having more queries)

metlos: I glossed over this when we were discussing your gist, but I don't know if this is really desirable.. Here you request all relationships with given name. Given we almost exclusively use this method to compute over `contains` relationships, this list is going to be absolutely huge (N less than the total number of entities in the whole inventory in all tenants, where N is the number of tenants). I understand the motivation of doing this with only a single query, but I am not sure if this is going to be feasible.

Maybe we could think about the "store whole subtree" idea circulated on the call to be used for this, too...

hawkular-inventory-impl-cassandra/src/main/java/org/hawkular/inventory/impl/cassandra/CassandraBackend.java
```diff
@@ -159,8 +170,59 @@ public CassandraBackend(RxSession session) {
...
+                    return session.execute(q);
+                })
+                .map(relationships -> computeTransitiveClosure(startingCP, relationships))
+                .flatMap(this::loadEntities);
+        return closureContent.toBlocking().getIterator();
+    }
+
+    private Observable<String> getRelationshipTables(Relationships.Direction direction) {
```
metlos: Is it necessary to wrap the table name in an observable here when you always do `Observable.just()`?

hawkular-inventory-impl-cassandra/src/main/java/org/hawkular/inventory/impl/cassandra/CassandraBackend.java
```diff
@@ -159,8 +170,59 @@ public CassandraBackend(RxSession session) {
...
+                throw new IllegalStateException("Unsupported direction: " + direction);
+        }
+    }
+
+    @VisibleForTesting
+    static Set<String> computeTransitiveClosure(String startingCP, Iterable<Row> relationships) {
+        // Breadth-first wanted: first-in-first-out and keep processed ordering
+        Deque<String> processFifo = new LinkedList<>();
+        Set<String> processed = new LinkedHashSet<>();
+        Multimap<String, Row> mapBySource = Multimaps.index(relationships, row -> row.getString(Statements.SOURCE_CP));
```
metlos: Does the `SOURCE_CP` work in both directions?

hawkular-inventory-impl-cassandra/src/main/java/org/hawkular/inventory/impl/cassandra/CassandraBackend.java
```diff
@@ -195,8 +257,15 @@ public boolean hasRelationship(Row entity, Relationships.Direction direction, St
 
     @Override
     public Set<Row> getRelationships(Row entity, Relationships.Direction direction, String... names) {
```
metlos, line 249: I know that the model changes a lot, but could you update this reflect the latest model, where relationships are modelled in 3 tables? One "main" table, and in/out tables that are meant used for hops and filtering.

hawkular-inventory-impl-cassandra/src/main/java/org/hawkular/inventory/impl/cassandra/Statements.java
```diff
@@ -36,6 +39,10 @@
     static final String TARGET_CP = "target_cp";
     static final String NAME = "name";
     static final String PROPERTIES = "properties";
+    static final Map<String, String> RELATIONSHIP_MAIN_CP = ImmutableMap.<String, String>builder()
```
metlos: This I think becomes unnecessary in the latest version of the model where we have 3 relationship tables.

hawkular-inventory-impl-cassandra/src/test/java/org/hawkular/inventory/impl/cassandra/CassandraBackendTest.java
```diff
@@ -0,0 +1,64 @@
...
+import com.google.common.collect.Lists;
+
+/**
+ * @author Joel Takvorian
+ */
+public class CassandraBackendTest {
+
+    @Test
+    public void shouldProcessTransitiveClosure() {
+        List<Row> relationships = Lists.newArrayList(
```
metlos: What's the performance characteristics of this with say 20000 relationships processed where only let's say 20 are in the resulting closure?

metlos (changes requested):

hawkular-inventory-impl-cassandra/src/main/java/org/hawkular/inventory/impl/cassandra/CassandraBackend.java
jotak: We need to create an observable pipeline soon or later, and it's a pretty way (well, "pretty" is subjective) to handle single table case (incoming and outgoing) and multiple tables case (both) in the same, general way. So there's no specific code for "both".

hawkular-inventory-impl-cassandra/src/main/java/org/hawkular/inventory/impl/cassandra/CassandraBackend.java
jotak: I guess not - btw I need to understand your new data model

hawkular-inventory-impl-cassandra/src/main/java/org/hawkular/inventory/impl/cassandra/CassandraBackend.java
jotak: Given what we said yesterday, would it make sense to add a feedId in the primary keys?

hawkular-inventory-impl-cassandra/src/main/java/org/hawkular/inventory/impl/cassandra/CassandraBackend.java
metlos: Agreed, let's keep it that way then...

hawkular-inventory-impl-cassandra/src/main/java/org/hawkular/inventory/impl/cassandra/CassandraBackend.java
metlos: We also compute the transitive closure when deleting stuff - when you delete an entity, you also delete everything it contains, so adding just the `feedId` wouldn't help in that case.

jotak:
@metlos I changed the closure algo so that it's not reading everything at once. Now it performs queries step by step, which is not necessarily what we want in the end but at least it should work for now.

I decoupled the closure algo so that we can change that easily later: all we'll have to do is to provide a function that gives relationships from a vertex (no matters if it reads from memory or from DB)

hawkular-inventory-impl-cassandra/src/test/java/org/hawkular/inventory/impl/cassandra/CassandraInventoryDBTest.java
```diff
@@ -0,0 +1,177 @@
...
+
+        Assert.assertTrue(BACKEND.hasRelationship(appleTree, apple, "contains"));
+        Assert.assertTrue(BACKEND.hasRelationship(appleTree, Relationships.Direction.outgoing, "contains"));
+        Assert.assertTrue(BACKEND.hasRelationship(apple, Relationships.Direction.incoming, "contains"));
+        Assert.assertTrue(BACKEND.hasRelationship(apple, Relationships.Direction.both, "contains"));
+    }
+
+    @Test
+    public void shouldNotHaveRelationship() {
+        Row apple = BACKEND.persist(
```
jotak, line 99: Note for self: clean up on tearDown

jotak:
@metlos , I rebased on your latest changes and had to squash the commits for that.

Some notes here:

- In relationship_in/out, I had to remove some part of the primary key that you added. I don't know if it was something necessary for what's you're doing, but in my case this PK was breaking my queries. Another option would be to add secondary indexes. I hope we can avoid creating yet another set of tables.

- In Statements class I created methods that directly return the Observable<Row>, I think it's better that at least this class manages statements bindings so that we can see both queries and bindings in the same class.

